PR #9, opened 2025-09-02
fix code format on main

chunyuan-w:
```sh
pre-commit run --all-files --show-diff-on-failure
```


.pre-commit-config.yaml
```diff
@@ -27,7 +27,7 @@ repos:
     hooks:
       - id: ruff
         args: [--select=F401, --fixable=F401]
-        files: ^(benchmark/|docs/|examples/)
+        files: ^(docs/|examples/)
```
chunyuan-w, line 30: In the SGLang main, `ruff` will check `sglang/benchmark` not `sglang/sgl-kernel/benchmark` https://github.com/sgl-project/sglang/blob/main/.pre-commit-config.yaml#L30

PR #13, opened 2025-09-24
Enable Model Config Input via a Centralized Parser in utils.py

DiweiSun:
This PR proposes to enhance the model configuration system by introducing a flexible and reusable mechanism to load model parameters — either from a Hugging Face model config (--model-name) or via manual CLI overrides — instead of relying on hardcoded values. The implementation will be centralized in a new utils.py module to enable consistent configuration handling across multiple benchmark and inference scripts.

This change enables users to:

✅ Benchmark models using real-world configurations (e.g., deepseek-ai/DeepSeek-R1)
✅ Override specific parameters (e.g., --num-experts 128)
✅ Avoid hardcoded test configurations
✅ Reuse config logic across tools

### DIFF
--- a/benchmark/bench_moe_topk_softmax.py
+++ b/benchmark/bench_moe_topk_softmax.py
@@ -1,77 +1,76 @@
-import itertools
+# benchmark_topk_softmax.py
+# Benchmark script for topk_softmax operator: compares VLLM vs SGLang implementations
+# Supports two modes:
+#   1. --model-name provided → load config from HF model
+#   2. No --model-name → use default hardcoded test configurations
 
+import itertools
 import torch
 import triton
+from utils import parse_args, get_model_config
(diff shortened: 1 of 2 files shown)


benchmark/bench_moe_topk_softmax.py
```diff
@@ -23,6 +24,22 @@ def vllm_topk_softmax(gating_output, topk):
     return topk_weights, topk_indices
 
 
+def navtive_topk_softmax(gating_output, topk):
```
chunyuan-w: This native function does not support `renormalize`. You may follow this one in SGLang:

https://github.com/sgl-project/sglang/blob/36942660513213f3ecd4a39ad64d6cf3127328a9/python/sglang/srt/layers/moe/topk.py#L411-L417

PR #54, opened 2025-12-09
Added support for sgl_per_tensor_quant_fp8

pralay-das:
This PR includes

- register sgl_per_tensor_quant_fp8 op in sgl kernel
- implement the op
- enable test case
-  fix lint issue in per_token_group_quant

To run the test case: `python -m pytest ./tests/test_per_tensor_quant_fp8.py`

Note: BMG currently uses cutlass FP8 emulation code. For CRI, when sycl compiler natively supports FP8 dtype, the code can use the exact FP8 dtypes directly with less modification.


src/sycl/per_tensor_quant_fp8.cpp
```diff
@@ -0,0 +1,236 @@
...
+    // Block-level max
+    auto work_group = item.get_group();
+    max_value = sycl::reduce_over_group(work_group, max_value, sycl::maximum<>());
+
+    // update global maximum scale
+    if (tid == 0) {
+      float local_scale = max_value / FP8_E4M3_MAX;
+      sycl::atomic_ref<
+          float,
+          sycl::memory_order::relaxed,
```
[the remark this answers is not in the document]
pralay-das: will discuss offline.

msinnha1 (approved):
lgtm!
Btw do we have any perf data with this kernel to set target/projections?

mingfeima:
@pralay-das do you have a roadmap for bigger scope, aka. what's next plan for this ? and what is this per tensor quant fp8 used in the model level?

mingfeima:
OK, I suppose you are referring fa8w4, correct me if i was wrong. Then what's the quantization recipe here? per tensor is not likely to work. Anyway, please give me a reference to prove that per tensor recipe works for DS 4bit weight only quantized model.

tests/test_per_tensor_quant_fp8.py
airMeng: don't forget to add it into ```run_suite.py```

src/sycl/per_tensor_quant_fp8.cpp
```diff
@@ -0,0 +1,235 @@
...
+      vec_type_in input_vec;
+      int64_t base_idx = i * VEC_SIZE;
+      input_vec.load(0, sycl::multi_ptr<const T, sycl::access::address_space::global_space>(input_ + base_idx));
+      vec_type_out output_vec;
+#pragma unroll
+      for (int j = 0; j < VEC_SIZE; ++j) {
+        float val = static_cast<float>(input_vec[j]) * scale_val;
+        val = sycl::fmax(-FP8_E4M3_MAX, sycl::fmin(val, FP8_E4M3_MAX));
+        DST_DTYPE fp8_val = static_cast<DST_DTYPE>(val);
+        output_vec[j] = sycl::bit_cast<output_storage_t>(fp8_val);
```
mingfeima: please check asm code to see if IGC has vectorized this.

src/sycl/per_tensor_quant_fp8.cpp
```diff
@@ -0,0 +1,235 @@
...
+    for (int64_t i = gid; i < num_vec_elems; i += grid_size) {
+      int64_t base_idx = i * VEC_SIZE;
+
+      vec_type input_vec;
+      input_vec.load(0, sycl::multi_ptr<const T, sycl::access::address_space::global_space>(input_ + base_idx));
+
+#pragma unroll
+      for (int j = 0; j < VEC_SIZE; ++j) {
+        float val = sycl::fabs(static_cast<float>(input_vec[j]));
+        max_value = sycl::fmax(max_value, val);
```
mingfeima: same as above.

src/sycl/per_tensor_quant_fp8.cpp
```diff
@@ -0,0 +1,235 @@
...
+      DISPATCH_VEC_SIZE(sycl::half, cutlass::float_e4m3_t, vec_size);
+    }
+  } else if (input.scalar_type() == at::ScalarType::BFloat16) {
+    if (output_q.scalar_type() == at::ScalarType::Float8_e4m3fn) {
+      int vec_size = preferred_vector_width(dpcppGetDeviceIdOfCurrentQueue(), sizeof(sycl::ext::oneapi::bfloat16));
+      DISPATCH_VEC_SIZE(sycl::ext::oneapi::bfloat16, cutlass::float_e4m3_t, vec_size);
+    }
+  } else {
+    throw std::runtime_error("Unsupported data type for per-tensor FP8 quantization");
+  }
```
mingfeima: use ` AT_DISPATCH_REDUCED_FLOATING_TYPES`

mingfeima: once you used  AT_DISPATCH_REDUCED_FLOATING_TYPES, you can just put `DISPATCH_VEC_SIZE` inside the function handler, since you will need to call it only once, no need to use MACRO.

mingfeima (dismissed):

mingfeima:
* the biggest issue is that this would be petty much dead code as no one uses per tensor fp8 quant recipe. so i suggest that your team has clear schedule of TODO list.
* other than that, you can simplify the code a little bit based on the comments.
* please benchmark your test code to see the achieved memory bandwitdh (this is also a prove that whether IGC does its job).

src/sycl/per_tensor_quant_fp8.cpp
sunjiweiswift: export IGC_ShaderDumpEnable=1
export IGC_DumpToCustomDir=./dump
Please set these two environment variables.

And build and run your case
There will be a dump file in your working directory. You need to check the simd16 assembly.

tests/test_per_tensor_quant_fp8.py
pralay-das: added.

src/sycl/per_tensor_quant_fp8.cpp
pralay-das: yes, I have dump the asm file, I am able to see 16|M0 , 32|M0 , 16|M16 , 1|M0 , 8|M0

kareemshaik80 (approved):

airMeng (approved):

deepvars (approved):

PR #159, opened 2026-04-09
Uncomment gelu config and fix typo in swiglu name

ck-intel:
Re-enable gelu config in bench_fused_moe.py which mistakenly got commented through PR #134 and fix a typo in the SWIGLU macro name.

### DIFF
--- a/benchmark/bench_fused_moe.py
+++ b/benchmark/bench_fused_moe.py
@@ -143,14 +143,14 @@ shape_configs = [
 
 shape_configs_gelu = [
     # grok, tp=1
-    # {
-    #     "num_experts": 8,
-    #     "topk": 2,
-    #     "hidden_size": 8192,
-    #     "shard_intermediate_size": 16384,
-    #     "dtype": torch.bfloat16,
-    #     "block_shape": None,
-    # }
+    {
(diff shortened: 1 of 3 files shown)


ck-intel:
@airMeng pls review. Thanks.

benchmark/bench_fused_moe.py
```diff
@@ -143,14 +143,14 @@
 
 shape_configs_gelu = [
```
airMeng, line 133: will you add configs for swiglu as well?

ck-intel: it's already there as part of the "shape_configs" list (at the end of that list as part of gpt-oss model config dict) defined just above "shape_configs_gelu" list.

airMeng: it might not be that explicit, can you move out as Gelu does?

ck-intel: I kept it like this as swiglu is just extended version of silu. I will update as suggested, will create one more list with name "shape_configs_swiglu_gpt_oss" and add gpt-oss related configs in that.

airMeng (approved):

PR #160, opened 2026-04-13
Fix CMake Incremental Build

sspintel:
- Remove unconditional object file deletion in run_sycl.cmake to enable incremental builds
- Auto-detect and use ccache/sccache for both C++ and SYCL compilations
- Persist FetchContent downloads outside build/ to avoid re-cloning cutlass
- Use stable build directory and --no-build-isolation for pip install

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,6 @@ build
 
 # vscode
 .vscode/
+
+# FetchContent persistent cache (cutlass, etc.)
+.deps/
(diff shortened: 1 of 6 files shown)


airMeng (approved):

sspintel:
Closing as the --no-build-isolation flag alone fixes the incremental build, and the CMake changes in this PR are not necessary. I will add that this flag to README in a later PR.

PR #172, opened 2026-04-21
MOE: shuffle kernel add fp8 support

kareemshaik80:


airMeng (commented):
no need to benchmark performance?

airMeng (approved):

PR #179, opened 2026-04-29
DSV4 MoE: add MXFP4 W4A16 support with staged weight dequantization

sspintel:
Introduces use_mxfp4_w4a16 mode for fused_experts on XPU, dequantizing w1/w2 just-in-time around each grouped GEMM to bound peak GPU memory.

What changed

- `python/sgl_kernel/moe.py`
  - Added `dequantize_mxfp4_weights()`: dequantizes OCP MXFP4 (E2M1 packed uint8 + UE8M0 block scales) to BF16 on-device using a lookup table.
  - Added `use_mxfp4_w4a16: bool` parameter to `fused_experts()`. When `True`, `w1`/`w2` are expected as packed `uint8` with corresponding `w1_scale`/`w2_scale` in UE8M0 format; input activations remain BF16 (W4A16).
  - Input validation and shape checks are adapted to account for the packed last dimension (`cols // 2`).

### DIFF
--- a/python/sgl_kernel/moe.py
+++ b/python/sgl_kernel/moe.py
@@ -4,10 +4,15 @@ import torch
 
 from .utils import is_xe2_arch
 
-# Lookup table for MXFP4 E2M1 dequantization: maps 3-bit magnitude codes to float values.
-# Index i corresponds to code i in {0,0.5,1.0,1.5,2.0,3.0,4.0,6.0}.
-_MXFP4_E2M1_TO_FLOAT = torch.tensor(
-    [0.0, 0.5, 1.0, 1.5, 2.0, 3.0, 4.0, 6.0], dtype=torch.float32
+# E2M1 lookup table: nibble value 0x0–0xF → float.
+# Encodes sign directly: 0b0xxx = positive, 0b1xxx = negative.
+# high nibble = second element; see _upcast_mxfp4_one_xpu in sglang).
+_MXFP4_E2M1_LUT = torch.tensor(
+    [
+        0.0, 0.5, 1.0, 1.5, 2.0, 3.0, 4.0, 6.0,        # 0b0xxx (positive)
+        0.0, -0.5, -1.0, -1.5, -2.0, -3.0, -4.0, -6.0,  # 0b1xxx (negative)
+    ],
+    dtype=torch.float32,
 )
 
 
@@ -257,10 +262,8 @@ def dequantize_mxfp4_weights(
     block_size: int = 32,
 ) -> torch.Tensor:
     """Dequantize MXFP4 (E2M1) packed expert weight tensor to BF16 (or other dtype).
-
-    MXFP4 uses the OCP MX format: 4-bit E2M1 data packed two-per-byte (low
-    nibble first), with one UE8M0 (uint8) block scale per ``block_size``
-    elements along the last dimension.
+        byte = (first_elem & 0xF) | (second_elem << 4)
+    i.e. low nibble = first element, high nibble = second element.
 
     Parameters:
     - packed:     [E, rows, packed_cols] uint8 – two FP4 nibbles per byte,
@@ -279,32 +282,25 @@ def dequantize_mxfp4_weights(
     cols = packed_cols * 2
     num_blocks = cols // block_size
 
-    # --- 1. Unpack two FP4 nibbles per byte (low nibble = element 0) ---
-    lo = packed & 0x0F  # [E, rows, packed_cols]
-    hi = (packed >> 4) & 0x0F  # [E, rows, packed_cols]
-    # Interleave lo/hi pairs → [E, rows, cols]
-    unpacked = torch.stack([lo, hi], dim=-1).reshape(E, rows, cols)
+    # --- 1. Unpack and dequantize via 16-entry signed LUT ---
+    # Low nibble = first element (even K position), high nibble = second element
+    lut = _MXFP4_E2M1_LUT.to(device=packed.device, dtype=torch.bfloat16)
+    lo = (packed & 0x0F).long()           # [E, rows, packed_cols]
+    hi = (packed >> 4).long()             # [E, rows, packed_cols]
+    # Interleave: [lo, hi] per packed position → [E, rows, cols]
+    paired = torch.stack([lut[lo], lut[hi]], dim=-1)  # [E, rows, packed_cols, 2]
+    dequantized = paired.reshape(E, rows, cols)
 
-    # --- 2. Dequantize E2M1 codes via lookup table ---
-    # E2M1 layout: bit 3 = sign, bits 2-0 = magnitude index
-    lut = _MXFP4_E2M1_TO_FLOAT.to(device=packed.device)
-    sign = ((unpacked >> 3) & 1).to(torch.bool)
-    magnitude_idx = (unpacked & 0x07).to(torch.long)
-    magnitude = lut[magnitude_idx]  # [E, rows, cols], float32
-    dequantized = torch.where(sign, -magnitude, magnitude)
-
-    # --- 3. Apply per-block UE8M0 scale factors ---
-    # Reshape into blocks: [E, rows, num_blocks, block_size]
-    dequantized_blocks = dequantized.reshape(E, rows, num_blocks, block_size)
-
-    # Decode UE8M0: float_scale = 2^(stored_byte - 127)
+    # --- 2. Apply per-block UE8M0 scale factors ---
+    dequantized = dequantized.reshape(E, rows, num_blocks, block_size)
     scale_exp = scales.to(torch.int32) - 127  # [E, rows, num_blocks]
-    # unsqueeze for broadcast over block_size dim
-    scale_values = torch.pow(2.0, scale_exp.float()).unsqueeze(-1)
+    scale_values = torch.exp2(scale_exp.float()).unsqueeze(-1).to(torch.bfloat16)
+    dequantized = dequantized * scale_values
 
-    scaled = dequantized_blocks * scale_values  # [E, rows, num_blocks, block_size]
-
-    return scaled.reshape(E, rows, cols).to(dtype)
+    result = dequantized.reshape(E, rows, cols)
+    if dtype != torch.bfloat16:
+        result = result.to(dtype)
+    return result
 
 
 def fused_experts(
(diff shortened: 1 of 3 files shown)


python/sgl_kernel/moe.py
```diff
@@ -244,6 +250,63 @@ def cutlass_fp4_group_mm(
...
+    - Tensor of shape [E, rows, cols] in ``dtype`` on the same device as
+      ``packed``.
+    """
+    E, rows, packed_cols = packed.shape
+    cols = packed_cols * 2
+    num_blocks = cols // block_size
+
+    # --- 1. Unpack two FP4 nibbles per byte (low nibble = element 0) ---
+    lo = packed & 0x0F  # [E, rows, packed_cols]
+    hi = (packed >> 4) & 0x0F  # [E, rows, packed_cols]
```
jayachandranb-ai: hi = (packed >> 4) & 0x0F 
shouldnt need & 0x0F .. packed is uint8 so right shift sould not sign extend.
after veryinfying existing code, you can remove the $0x0F and see if things work ok

python/sgl_kernel/moe.py
```diff
@@ -244,6 +250,63 @@ def cutlass_fp4_group_mm(
...
+    # Interleave lo/hi pairs → [E, rows, cols]
+    unpacked = torch.stack([lo, hi], dim=-1).reshape(E, rows, cols)
+
+    # --- 2. Dequantize E2M1 codes via lookup table ---
+    # E2M1 layout: bit 3 = sign, bits 2-0 = magnitude index
+    lut = _MXFP4_E2M1_TO_FLOAT.to(device=packed.device)
```
jayachandranb-ai: you can use a table containing -ve values as well and do a straight lookup using "unpacked":
float values from 0000 to 1111, are:
[0.0, 0.5, 1.0, 1.5, 2.0, 3.0, 4.0, 6.0, -0.0, -0.5, -1.0, -1.5, -2.0, -3.0, -4.0, -6.0]

note bit pattern 0000 maps to 0.0 and 1000 maps to -0.0

python/sgl_kernel/moe.py
```diff
@@ -244,6 +250,63 @@ def cutlass_fp4_group_mm(
...
+    dequantized = torch.where(sign, -magnitude, magnitude)
+
+    # --- 3. Apply per-block UE8M0 scale factors ---
+    # Reshape into blocks: [E, rows, num_blocks, block_size]
+    dequantized_blocks = dequantized.reshape(E, rows, num_blocks, block_size)
+
+    # Decode UE8M0: float_scale = 2^(stored_byte - 127)
+    scale_exp = scales.to(torch.int32) - 127  # [E, rows, num_blocks]
+    # unsqueeze for broadcast over block_size dim
+    scale_values = torch.pow(2.0, scale_exp.float()).unsqueeze(-1)
```
jayachandranb-ai: using torch.exp2(scale_exp) should be more optimal. it can take input as integer so no need of scale_exp.float() cast

python/sgl_kernel/moe.py
```diff
@@ -315,10 +383,34 @@ def fused_experts(
...
+        assert (
+            w1_scale is not None
+        ), "w1_scale (UE8M0 uint8) must be provided when use_mxfp4_w4a16=True"
+        assert (
+            w2_scale is not None
+        ), "w2_scale (UE8M0 uint8) must be provided when use_mxfp4_w4a16=True"
+        assert w1_scale.dtype == torch.uint8, "w1_scale must be uint8 (UE8M0 format)"
+        assert w2_scale.dtype == torch.uint8, "w2_scale must be uint8 (UE8M0 format)"
+    else:
+        assert w1_scale is None, "current MoE does not support w1_scale"
```
jayachandranb-ai: message can be modified as 
current MoE supports w1_scale only if  use_mxfp4_w4a16 is true.
similarly for w2

python/sgl_kernel/moe.py
```diff
@@ -349,6 +445,10 @@ def fused_experts(
 
     E, _, K = w1.shape
     E, OutK, N = w2.shape
+    if use_mxfp4_w4a16:
+        # w1/w2 last dims are packed (H//2, I//2); recover actual dims
+        K = K * 2
+        N = N * 2
     assert N * 2 == w1.shape[1], "w1 shape[1] must be 2x of w2 shape[2]"
```
jayachandranb-ai: may be we can keep this assert and remove the similar assert done earlier or vice versa.. not immediately needed..

python/sgl_kernel/moe.py
sspintel: Done

python/sgl_kernel/moe.py
sspintel: Done. Include a 16 value LUT with -ve values.

python/sgl_kernel/moe.py
sspintel: Done.

python/sgl_kernel/moe.py
sspintel: Done.

python/sgl_kernel/moe.py
```diff
@@ -467,6 +577,8 @@ def fused_experts(
             gemm1_alpha=float(gemm1_alpha) if gemm1_alpha is not None else 1.702,
             gemm1_limit=float(gemm1_limit) if gemm1_limit is not None else 7.0,
         )
+        if use_mxfp4_w4a16:
+            del w2
```
jayachandranb-ai, line 637: is  torch.xpu.empty_cache() also needed?

python/sgl_kernel/moe.py
```diff
@@ -426,6 +526,10 @@ def fused_experts(
     avg_m = (M * TopK) // E
     big_weight = K * N > 4096 * 4096
     use_unfused_act = avg_m <= 128 and big_weight
+    # ---- GEMM1: dequantize w1 just before use, free immediately after ----
+    if use_mxfp4_w4a16:
```
jayachandranb-ai: For better code readability, shall we do like this ( dequant immediately above Gemm, remove dequant gemm immediately after gemm )

if use_unfused_act:
.......
       if use_mxfp4_w4a16:
              w1 = dequant
       torch.ops.sgl_kernel.moe_grouped_mm_nt_xe20()
       if use_mxfp4_w4a16:
               del w1 and free cache

 ..........................
      if use_mxfp4_w4a16:
              w2 = dequant
       torch.ops.sgl_kernel.moe_grouped_mm_nt_xe20()
       if use_mxfp4_w4a16:
               del w2 and free cache

else:

...........
 similar code as above..



Currently, w1 quant is done outside if use_unfused_act:  While it reduces some code, it may be a bit unreadable.

python/sgl_kernel/moe.py
```diff
@@ -244,6 +255,54 @@ def cutlass_fp4_group_mm(
...
+    lo = (packed & 0x0F).long()           # [E, rows, packed_cols]
+    hi = (packed >> 4).long()             # [E, rows, packed_cols]
+    # Interleave: [lo, hi] per packed position → [E, rows, cols]
+    paired = torch.stack([lut[lo], lut[hi]], dim=-1)  # [E, rows, packed_cols, 2]
+    dequantized = paired.reshape(E, rows, cols)
+
+    # --- 2. Apply per-block UE8M0 scale factors ---
+    dequantized = dequantized.reshape(E, rows, num_blocks, block_size)
+    scale_exp = scales.to(torch.int32) - 127  # [E, rows, num_blocks]
+    scale_values = torch.exp2(scale_exp.float()).unsqueeze(-1).to(torch.bfloat16)
```
jayachandranb-ai: scale_exp.float() should not be needed.   scale_exp may work .( it should as per pytorch i.e, int should be supported as argument py by exp2).

python/sgl_kernel/moe.py
```diff
@@ -244,6 +255,54 @@ def cutlass_fp4_group_mm(
...
+      ``packed``.
+    """
+    E, rows, packed_cols = packed.shape
+    cols = packed_cols * 2
+    num_blocks = cols // block_size
+
+    # --- 1. Unpack and dequantize via 16-entry signed LUT ---
+    # Low nibble = first element (even K position), high nibble = second element
+    lut = _MXFP4_E2M1_LUT.to(device=packed.device, dtype=torch.bfloat16)
+    lo = (packed & 0x0F).long()           # [E, rows, packed_cols]
```
jayachandranb-ai: is .long() needed? lo can be  int32 or even uint8. right?

python/sgl_kernel/moe.py
```diff
@@ -255,6 +314,7 @@ def fused_experts(
     inplace: bool = False,
     activation: str = "silu",
     use_fp8_w8a8: bool = False,
+    use_mxfp4_w4a16: bool = False,
```
jayachandranb-ai, line 379: This is an api change. Let us check with Workloads(Mohit) if this change is ok. (this is the clean way, just that sglang repo will need a change to add this api param. ) Otherwise we need to detect the type of w1 / w2/ scales and if they are  uint8 assume it is mxfp4  and set he flag use_mxfp4_w4a16 internally at the top. ( which is dirty)

rahulvijayaraghavan:
Please check if optimizations here can be incorporated:
https://github.com/rahulvijayaraghavan/sglang/commit/51ff7970747e1b307d991375300c1bad1170b754

python/sgl_kernel/moe.py
```diff
@@ -315,10 +451,34 @@ def fused_experts(
...
+    # For MXFP4 W4A16: validate packed uint8 inputs and scales.
+    # Actual dequantization is deferred to just before each GeMM so that
+    # at most one dequantized weight tensor lives in GPU memory at a time.
+    # Scales must be None on all non-mxfp4 code paths.
+    if use_mxfp4_w4a16:
+        assert (
+            w1.dtype == torch.uint8
+        ), "use_mxfp4_w4a16=True requires w1 to be uint8 (packed MXFP4)"
+        assert (
+            w2.dtype == torch.uint8
```
polisettyvarma: it is int8

python/sgl_kernel/moe.py
```diff
@@ -315,10 +451,34 @@ def fused_experts(
...
+        ), "use_mxfp4_w4a16=True requires w2 to be uint8 (packed MXFP4)"
+        assert (
+            w1_scale is not None
+        ), "w1_scale (UE8M0 uint8) must be provided when use_mxfp4_w4a16=True"
+        assert (
+            w2_scale is not None
+        ), "w2_scale (UE8M0 uint8) must be provided when use_mxfp4_w4a16=True"
+        assert w1_scale.dtype == torch.uint8, "w1_scale must be uint8 (UE8M0 format)"
+        assert w2_scale.dtype == torch.uint8, "w2_scale must be uint8 (UE8M0 format)"
```
polisettyvarma: fp32 scales are sent directly
change assert and remove this conversion everywhere

polisettyvarma (changes requested):

sspintel:
Closing as this was a stop-gap workaround to the eventual tile-fused kernel which gives more performance.

PR #184, opened 2026-05-06
  DSV4: MXFP4 W4A16 tile-fused grouped GEMM for BMG

sspintel:
### Summary
    
This PR brings up a tile-fused MXFP4 W4A16 grouped GEMM on Intel BMG (Xe2) and wires it end-to-end through the MoE fused_experts path. Dequantization of MXFP4 weights is fused into the GEMM mainloop with collaborative scale loading, replacing the prior W4A16 dequant-then-GEMM pipeline.

###  What this PR does

- Tile-fused MXFP4 mainloop on Xe2: a new MoEMainloopMxfp4W4A16 (src/sycl/kernels/moe/xe20/mxfp4_w4a16/moe_mainloop.hpp) consumes packed FP4 weights, applies per-block FP32 scales (one scale per 32 K-elements) directly on the MMA-B fragment via apply_B_scales_mma, and feeds bf16 operands into the matrix engines. Scales are loaded collaboratively per subgroup and reused across the K-loop.
- End-to-end wiring: fused_experts calls moe_grouped_mm_nt_xe20_mxfp4_w4a16 directly; Scales use the N-outer [E, N, K/32] layout.
- Reorganized Xe2 MoE kernel headers: split into bf16/, mxfp4_w4a16/, and a new common/ directory. Fused gate-and-mul activations (SILU / GELU / SWIGLU_GPT_OSS) now live in common/activation.hpp and are shared by both mainloops, so future activations (e.g. ReLU²) drop into one file instead of two.
- Shared dispatch heuristic: MOE_GROUPED_GEMM_SMALL_WEIGHT_THRESHOLD (4096×4096) is centralized in src/sycl/Utils.h and used by both GroupGemmXe20.cpp and GroupGemmMxfp4W4A16Xe20.cpp for the small-vs-big-weight tile selection. The Python fused_experts activation heuristic references the same constant and documents the requirement to keep them in sync.
- Pruned AOT instantiation matrix: the MXFP4 fused kernel is built for ActType=SILU and WithBias=false only (for DSV4). Other combos TORCH_CHECK-fail with a message pointing at src/GroupGemmMxfp4W4A16Xe20.cmake to re-enable them. This is a deliberate trade-off to keep Level Zero module pressure within budget under TP>1.
- Minor mainloop cleanup: hoisted loop-invariant sg_n_coord / n_sg_base out of the K-loop in both gated and split (gate/up) paths.


src/torch_extension_sycl.cc
```diff
@@ -95,47 +98,6 @@ TORCH_LIBRARY_FRAGMENT(sgl_kernel, m) {
```
jayachandranb-ai, line 104: why was this code removed?

sspintel: That was a mistake. I had removed a few other kernels to accelerate build time. Will add them back.

src/sycl/kernels/moe/xe20/mxfp4_w4a16/moe_kernel.hpp
airMeng, line 1: How about ```src/sycl/kernels/moe/xe20/mxfp4/moe_mxfp4_kernel.hpp```, and you can rename the bf16 files as ```src/sycl/kernels/moe/xe20/bf16/moe_kernel.hpp```

airMeng: BTW, I think ```src/sycl/kernels/moe/xe20_mxfp4/moe_mxfp4_kernel.hpp``` is basically the same as bf16 version, could you extract them into the same templates?

src/sycl/kernels/moe/xe20/mxfp4_w4a16/moe_mainloop.hpp
```diff
@@ -0,0 +1,558 @@
...
+#include "cutlass/util/reference/device/tensor_compare.h"
+#include "cutlass/util/reference/host/tensor_fill.h"
+#include "cutlass/util/sycl_event_manager.hpp"
+
+#pragma clang diagnostic ignored "-Wpass-failed"
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+
+#define SILU 0
+#define GELU 1
+#define SWIGLU_GPT_OSS 2
```
airMeng: we might reuse macros in bf16 files, for example, extract them to some common files. Then we can save effort if we need to add new activations like https://github.com/sgl-project/sgl-kernel-xpu/pull/194

src/sycl/kernels/moe/xe20/mxfp4_w4a16/moe_mainloop.hpp
```diff
@@ -0,0 +1,558 @@
...
+    scale_bf16[i] = cutlass::bfloat16_t(scales_sg[i]);
+  }
+
+  CUTE_UNROLL
+  for (int idx = 0; idx < frag_size; ++idx) {
+    auto coord = coord_frag(idx);
+    int n_in_sg = get<0>(coord) - n_sg_base;
+    cutlass::bfloat16_t s = scale_bf16[n_in_sg];
+    frag(idx) = frag(idx) * s;
+  }
```
airMeng, line 155: I wonder whether exponent manipulation will be faster since the scales of mxfp4 are barely the power of 2?

src/sycl/GroupGemmMxfp4W4A16Xe20.cpp
```diff
@@ -0,0 +1,193 @@
...
+using Tile_8_64_32 = Shape<_8, _64, _32>;
+using Tile_128_64_32 = Shape<_128, _64, _32>;
+using Tile_128_128_32 = Shape<_128, _128, _32>;
+using Tile_256_64_32 = Shape<_256, _64, _32>;
+using Tile_256_256_32 = Shape<_256, _256, _32>;
+
+using SG_1_4_1 = Layout<Shape<_1, _4, _1>, Stride<_4, _1, _0>>;
+using SG_4_2_1 = Layout<Shape<_4, _2, _1>, Stride<_2, _1, _0>>;
+using SG_8_2_1 = Layout<Shape<_8, _2, _1>, Stride<_2, _1, _0>>;
+using SG_8_4_1 = Layout<Shape<_8, _4, _1>, Stride<_4, _1, _0>>;
```
airMeng, line 47: have you tuned the configs here? BF16 uses these configs based on tuning, you might not need to follow

src/sycl/GroupGemmMxfp4W4A16Xe20.cpp
```diff
@@ -0,0 +1,193 @@
...
+  at::Tensor atomic_buffer = at::empty({static_cast<long>(1)}, activations.options().dtype(at::kInt));
+  bool with_bias = bias.has_value();
+  void* bias_ptr = with_bias ? bias->data_ptr() : nullptr;
+
+  int avg_m = total_m / n_experts;
+  bool small_weight = (int64_t)gemm_k * gemm_n <= (int64_t)4096 * 4096;
+
+  if (avg_m <= 8) {
+    DISPATCH_MOE_MXFP4(
+        activation_type, fuse_act, with_bias, Shape<_8, _64, _32>, Layout<Shape<_1, _4, _1>, Stride<_4, _1, _0>>);
```
airMeng, line 195: same as above

airMeng (commented):
Overall GREAT WORK! Do you have any performance data to show?

src/sycl/GroupGemmMxfp4W4A16Xe20.cpp
airMeng: I mean if the performance optimization is still ongoing, you can shrink configurations to lower the impact on binary size. You can keep only 1 set of configs if this PR is mostly for functionality

kareemshaik80 (commented):
overall LGTM.

src/sycl/GroupGemmMxfp4W4A16Xe20.cpp
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  auto stream = at::xpu::getCurrentXPUStream();
+  auto queue = stream.queue();
+  at::Tensor atomic_buffer = at::empty({static_cast<long>(1)}, activations.options().dtype(at::kInt));
+  bool with_bias = bias.has_value();
+  void* bias_ptr = with_bias ? bias->data_ptr() : nullptr;
+
+  int avg_m = total_m / n_experts;
+  bool small_weight = (int64_t)gemm_k * gemm_n <= (int64_t)4096 * 4096;
```
kareemshaik80: You can extract the threshold 4096*4096 in one place, it affects on c++ dispatch Heuristic.  Change in one place breaks the other.

src/sycl/kernels/moe/xe20/mxfp4_w4a16/moe_mainloop.hpp
```diff
@@ -0,0 +1,558 @@
...
+      reorder(tArA, tSrA);
+      // Fused type-conversion + layout-reorder. This picks the
+      // ConvertRelayout dispatch inside cute::reorder, which handles both
+      // E2M1 → bf16 upcast and copy-layout → MMA-B layout permutation.
+      reorder(tBrB_packed, tSrB);
+
+      // Apply per-block MXFP4 scales to tSrB in-place. Each element's
+      // N-coord is looked up in the MMA-B coord companion; scale_sg holds
+      // the per-SG fp32 scales are cast to bf16 in apply_B_scales_mma.
+      const int sg_n_coord = (thr_id / SUBGROUP_SIZE) % ATOM_N_V;
```
kareemshaik80: This is recomputed for every k-tile, move it outside loop.

src/sycl/GroupGemmMxfp4W4A16Xe20.cpp
```diff
@@ -0,0 +1,193 @@
...
+      static_cast<float>(gemm1_limit))
+
+// TEMPORARY: matching prune for the cmake-side instantiation matrix above.
+// Only ActType=0 (silu) and WithBias=false are instantiated, so the dispatch
+// hard-codes those values rather than template-branching on runtime inputs.
+// Callers that need the other combos will hit the TORCH_CHECK below and must
+// restore the full matrix before rebuilding.
+#define DISPATCH_MOE_MXFP4(ActType, FuseAct, WithBias, ...)                                                      \
+  do {                                                                                                           \
+    TORCH_CHECK((ActType) == 0, "mxfp4 fused kernel built with ActType=0 only (silu); got ActType=", (ActType)); \
```
kareemshaik80: if GELU and other activations are not fused we can remove in the PR description.

src/sycl/GroupGemmMxfp4W4A16Xe20.cpp
```diff
@@ -0,0 +1,193 @@
...
+    TORCH_CHECK(                                                                                                 \
+        !(WithBias), "mxfp4 fused kernel built with WithBias=false only; bias path unsupported in this build");  \
```
kareemshaik80: error messages don't explain why (L0 module pressure). you can add context for the error message matching cmake config. Minor commnet.

src/sycl/kernels/moe/xe20/mxfp4_w4a16/moe_kernel.hpp
sspintel: I have addressed point 1 and move the kernle files to their respective mxfp4/bf16 folders. Although most of the change in the kernel files remain the same, Point 2 is a bit tricky to handle, so I am skipping it for now in favor of maintainability.

src/sycl/kernels/moe/xe20/mxfp4_w4a16/moe_mainloop.hpp
sspintel: Good suggestion, Done. Added a new activations.h header with activations shared between both mxfp4 and bf16 files.

src/sycl/GroupGemmMxfp4W4A16Xe20.cpp
sspintel: Actually, I have used the configs that work well for the mxfp4 kernel also with good improvement. I have also drastically reduced the # of configurations by only keeping the activation + bias config supported by DSV4 for now.

src/sycl/GroupGemmMxfp4W4A16Xe20.cpp
sspintel: Done, added a flag that is shared in both places where the threshold calc is used.

src/sycl/kernels/moe/xe20/mxfp4_w4a16/moe_mainloop.hpp
sspintel: Done, moved out of loop in all occurrences.

src/sycl/GroupGemmMxfp4W4A16Xe20.cpp
sspintel: Done.

src/sycl/GroupGemmMxfp4W4A16Xe20.cpp
sspintel: Fixed, made the error messages more verbose.

src/sycl/kernels/moe/xe20/mxfp4_w4a16/moe_mainloop.hpp
sspintel: In the flow of DSV4 mxfp4 expert weight scales, they are decoded on the host before launch of the kernel, so they are received as fp32 instead of ue8m0. Worth attempting in a follow-up change if it's required and provides a resonable improvement.

airMeng (approved):

kareemshaik80 (approved):
LGTM

PR #203, opened 2026-05-18
[Norm] enable different data types of input and weight

airMeng:
In gemma4, weight tensors are always f32. The PR infactor norm kernels to accept f32 weight natively.


gaopengff:
LGTM

gaopengff (approved):

PR #205, opened 2026-05-19
enable fused RELU2 for non-gated architecture, to optimize performance

yao-matrix:


python/sgl_kernel/moe.py
```diff
@@ -343,13 +343,17 @@ def fused_experts(
         if is_xe2_arch() and b1.dtype == torch.bfloat16:
             # cast b1 to float32, since bias is accumulated in float32 in the kernel
             b1 = b1.float()
+            # Ensure conversion is complete before proceeding
+            torch.xpu.synchronize()
```
airMeng: remove the forced synchronization to avoid host overhead

python/sgl_kernel/moe.py
```diff
@@ -495,6 +495,7 @@ def fused_experts(
             gemm1_limit=float(gemm1_limit) if gemm1_limit is not None else 7.0,
         )
     else:
+        # Use zeros instead of empty to avoid uninitialized memory issues
```
airMeng: useless

src/GroupGemmXe20.cmake
```diff
@@ -16,13 +16,8 @@ endfunction()
...
     foreach(with_bias true false)
-        # RELU2 (act_type=3) only supports fuse_act=false (unfused activation)
-        # Skip fuse_act=true instantiations for act_type=3 to reduce binary size
-        if(act_type EQUAL 3)
-            set(fuse_act_list false)
-        else()
-            set(fuse_act_list true false)
-        endif()
+        # All activation types now support both fused and unfused paths
+        set(fuse_act_list true false)
```
airMeng, line 25: I think nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16 doesn't need bias, remove to shrink binary size

src/sycl/GroupGemmXe20.cpp
```diff
@@ -205,7 +206,7 @@ void moe_grouped_mm_nt_xe20(
   TORCH_CHECK(
       activation_type >= static_cast<int>(ActivationType::MIN) &&
           activation_type <= static_cast<int>(ActivationType::MAX),
-      "Unsupported activation_type: ",
+      "Unsupported activation_typ: ",
```
airMeng:
```suggestion
      "Unsupported activation_type: ",
```

yao-matrix:
@airMeng resolved comments, pls help review again, thx

src/GroupGemmXe20.cmake
airMeng, line 25: I mean ``` if act_type==3: set(with_bias false)``` does it make sense?

airMeng (commented):
Except there shall be no combination of relu and bias=true, others LGTM

airMeng (approved):

PR #225, opened 2026-05-28
Add fused RoPE + KV-cache write kernel

jmunetong:
## Summary

Add a SYCL kernel that applies rotary position embedding (RoPE) to Q/K in-place and writes rotated K + V into the flat KV-cache — all in a single kernel launch. Replaces 3 separate launches (rotary_embedding + 2x index_put) with 1 for eligible layers.

## Changes

- `src/sycl/Rope.cpp`: new `FusedRopeKVCacheKernel` + `apply_rope_inplace_with_kvcache_xpu` (~194 lines)
- `include/sgl_kernel_ops.h`: forward declaration
- `src/torch_extension_sycl.cc`: op registration under `sgl_kernel` TORCH_LIBRARY_FRAGMENT
- `python/sgl_kernel/elementwise.py`: Python wrapper with docstring
- `python/sgl_kernel/__init__.py`: export

## Kernel details

- One work-group per token, work-items stride over `(heads × rotary_dim/2)` pairs
- fp32 cos_sin_cache arithmetic (upcast from bf16 Q/K, downcast result) for numerical stability
- Handles both NeoX (rotate-half) and interleaved RoPE styles
- Skips tokens with `out_loc < 0` (speculative decoding)
- Handles `head_dim > rot_dim` by copying the non-rotated K tail to cache
- Mirrors the CUDA `apply_rope_inplace_with_kvcache` from `sglang/jit_kernel/rope.py`

## Companion sglang PR

- [sglang #26595](https://github.com/sgl-project/sglang/pull/26595): wires `forward_xpu` to dispatch to this kernel via `fused_set_kv_buffer_arg`

## Test

- 5/5 bit-parity unit tests pass (`test_rope_kvcache_fused.py`)
- Verified: fused output matches unfused reference (RoPE + explicit cache write) within `rtol=1e-2, atol=1e-3` on bf16
- Note: model-level integration (Part B in gemma4_causal.py) is currently disabled pending investigation of chunked-prefill interaction with SWAKVPool

### DIFF
--- a/include/sgl_kernel_ops.h
+++ b/include/sgl_kernel_ops.h
@@ -190,7 +190,7 @@ void fused_qk_rope_with_cos_sin_cache_inplace(
     bool is_neox);
 void sgl_per_token_group_quant_fp4(
     at::Tensor input, at::Tensor output_q, at::Tensor output_s, int64_t group_size, double eps);
-void apply_rope_inplace_with_kvcache_xpu(
+void apply_rope_inplace_with_kvcache(
     at::Tensor& query,
     at::Tensor& key,
     at::Tensor& value,
--- a/include/sgl_kernel_ops.h
+++ b/include/sgl_kernel_ops.h
@@ -190,6 +190,16 @@ void fused_qk_rope_with_cos_sin_cache_inplace(
     bool is_neox);
 void sgl_per_token_group_quant_fp4(
     at::Tensor input, at::Tensor output_q, at::Tensor output_s, int64_t group_size, double eps);
+void apply_rope_inplace_with_kvcache_xpu(
+    at::Tensor& query,
+    at::Tensor& key,
+    at::Tensor& value,
+    at::Tensor& k_cache,
+    at::Tensor& v_cache,
+    at::Tensor& cos_sin_cache,
+    at::Tensor& positions,
+    at::Tensor& out_loc,
+    bool is_neox);
 }  // namespace at::native::xpu
 void silu_and_mul(torch::Tensor& out, torch::Tensor& input);
 void gelu_tanh_and_mul(torch::Tensor& out, torch::Tensor& input);
--- a/python/sgl_kernel/elementwise.py
+++ b/python/sgl_kernel/elementwise.py
@@ -195,7 +195,7 @@ def gelu_and_mul(input: torch.Tensor, out: torch.Tensor = None) -> torch.Tensor:
     return out
 
 
-def apply_rope_inplace_with_kvcache_xpu(
+def apply_rope_inplace_with_kvcache(
     query: torch.Tensor,
     key: torch.Tensor,
     value: torch.Tensor,
@@ -238,7 +238,7 @@ def apply_rope_inplace_with_kvcache_xpu(
     if cos_sin_cache.dtype != torch.float32:
         raise ValueError("cos_sin_cache must be float32")
 
-    torch.ops.sgl_kernel.apply_rope_inplace_with_kvcache_xpu(
+    torch.ops.sgl_kernel.apply_rope_inplace_with_kvcache(
         query,
         key,
         value,
--- a/python/sgl_kernel/elementwise.py
+++ b/python/sgl_kernel/elementwise.py
@@ -195,6 +195,62 @@ def gelu_and_mul(input: torch.Tensor, out: torch.Tensor = None) -> torch.Tensor:
     return out
 
 
+def apply_rope_inplace_with_kvcache_xpu(
+    query: torch.Tensor,
+    key: torch.Tensor,
+    value: torch.Tensor,
+    k_cache: torch.Tensor,
+    v_cache: torch.Tensor,
+    cos_sin_cache: torch.Tensor,
+    positions: torch.Tensor,
+    out_loc: torch.Tensor,
+    is_neox: bool = True,
+) -> None:
+    r"""Fused in-place RoPE + KV-cache store for XPU.
+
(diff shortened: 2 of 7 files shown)


python/sgl_kernel/elementwise.py
```diff
@@ -195,6 +195,62 @@ def gelu_and_mul(input: torch.Tensor, out: torch.Tensor = None) -> torch.Tensor:
...
+        Must be float32.
+    positions : torch.Tensor
+        Position indices, shape: ``(num_tokens,)``, dtype int64.
+    out_loc : torch.Tensor
+        Flat cache slot indices, shape: ``(num_tokens,)``, dtype int64.
+        Use -1 to skip a token (speculative decoding).
+    is_neox : bool
+        Whether to use GPT-NeoX style (True) or interleaved (False).
+    """
+    if cos_sin_cache.dtype != torch.float32:
```
gaopengff, line 238: Currently sglang will cast **cos_sin_cache** to query datatype(bfloat16/half) for non-cuda device. Other datatypes for cache are allowed here.
https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/layers/rotary_embedding/base.py#L73

src/sycl/Rope.cpp
```diff
@@ -473,4 +473,203 @@ std::tuple<at::Tensor, at::Tensor> rotary_embedding(
...
+// ============================================================================
+// Fused RoPE + KV-cache write kernel
+// Mirrors CUDA `apply_rope_inplace_with_kvcache` from jit_kernel/rope.py:140-175.
+// Single kernel: rotates Q/K in-place (fp32 arithmetic) and writes rotated K + V
+// into the flat KV-cache at out_loc.
+// ============================================================================
+
+template <typename scalar_t, EmbeddingAlgorithm algo>
+struct FusedRopeKVCacheKernel {
+  using accscalar = at::opmath_type<scalar_t>;
```
gaopengff, line 510: For bfloat16/half, it's better not to case fp32. Calculating directly is OK as the cos_sin_cache is not guaranteed to be  fp32 for xpu device.

include/sgl_kernel_ops.h
```diff
@@ -190,6 +190,16 @@ void fused_qk_rope_with_cos_sin_cache_inplace(
     bool is_neox);
 void sgl_per_token_group_quant_fp4(
     at::Tensor input, at::Tensor output_q, at::Tensor output_s, int64_t group_size, double eps);
+void apply_rope_inplace_with_kvcache_xpu(
```
gaopengff: For kernel name, remove the xpu suffix. I think "apply_rope_inplace_with_kvcache" is OK here.

gaopengff (commented):
Please add unit test.

python/sgl_kernel/elementwise.py
gaopengff: I saw you sglang PR. So cos_sin_cache should be fp32 for xpu device in the future.

src/sycl/Rope.cpp
```diff
@@ -473,4 +473,203 @@ std::tuple<at::Tensor, at::Tensor> rotary_embedding(
...
 }
 
+// ============================================================================
+// Fused RoPE + KV-cache write kernel
+// Mirrors CUDA `apply_rope_inplace_with_kvcache` from jit_kernel/rope.py:140-175.
+// Single kernel: rotates Q/K in-place (fp32 arithmetic) and writes rotated K + V
+// into the flat KV-cache at out_loc.
+// ============================================================================
+
+template <typename scalar_t, EmbeddingAlgorithm algo>
```
gaopengff: Do you have plan to optimize this kernel? Looks like we could make **rope_dim**  be constant while launching kernel just like cuda jit kernel did.

src/sycl/Rope.cpp
```diff
@@ -473,4 +491,376 @@ std::tuple<at::Tensor, at::Tensor> rotary_embedding(
...
+                sycl::nd_range<1>(sycl::range<1>(num_tokens * group_size), sycl::range<1>(group_size)), kernel);
+          };
+          dpcppGetCurrentQueue().submit(cgf);
+        };
+
+        // Dispatch on the runtime rot_dim to a compile-time specialization when it is one
+        // of the common values, else fall back to the runtime-rot_dim kernel (ROT_DIM=0).
+        // 5 dims x 2 algos = 10 specializations per dtype, matching the DSRotaryEmbedding
+        // precedent's instantiation count.
+        auto dispatch_algo = [&](auto algo_tag) {
```
gaopengff, line 835: Better to use macro to launch kernel, which is more clear.

src/sycl/Rope.cpp
```diff
@@ -473,4 +491,376 @@ std::tuple<at::Tensor, at::Tensor> rotary_embedding(
...
+              launch(algo_tag, std::integral_constant<int, 96>{});
+              break;
+            case 128:
+              launch(algo_tag, std::integral_constant<int, 128>{});
+              break;
+            case 256:
+              launch(algo_tag, std::integral_constant<int, 256>{});
+              break;
+            default:
+              launch(algo_tag, std::integral_constant<int, 0>{});  // runtime-rot_dim fallback
```
gaopengff, line 853: Looks like in sglang cuda only supports [64, 128, 256, 512] in this kernel. Do we have scenario to use this runtime-rot-dim pass? If not supporting specific dims is enough, and we could unroll the dim related loop to optimize this kernel.
https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/layers/rotary_embedding/base.py#L79

PR #234, opened 2026-06-04
Support fp32 gating logits in topk_sigmoid

jmunetong:
Widen the topk_sigmoid dispatch from AT_DISPATCH_REDUCED_FLOATING_TYPES (Half, BFloat16 only) to AT_DISPATCH_FLOATING_TYPES_AND2 so float32 router gating logits (e.g. Nemotron-3-Nano MoE) are accepted. The kernel already upcasts each element to float internally, so the fp32 path is numerically a no-op cast.

Add test_topk_sigmoid_fp32 exercising the new fp32 path against the float reference across the same token/expert/topk/renormalize/bias grid.

### DIFF
--- a/tests/test_topk_sigmoid.py
+++ b/tests/test_topk_sigmoid.py
@@ -41,7 +41,13 @@ def fused_topk_sigmoid_torch_native(
     return topk_weights, topk_ids
 
 
-@pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float16])
+# float32 covers the fp32 gating-logits path: some models (e.g.
+# Nemotron-3-Nano MoE) emit float32 router logits. Before the kernel dispatch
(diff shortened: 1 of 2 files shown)


tests/test_topk_sigmoid.py
```diff
@@ -98,5 +98,62 @@ def test_topk_sigmoid(
     torch.testing.assert_close(res, ref, atol=atol, rtol=rtol)
 
 
+@pytest.mark.parametrize("n_token", [2, 32, 4096])
```
gaopengff: Better to reuse method 'test_topk_sigmoid' for fp32 test. You can add datatype with the following code:
"@pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float16, **torch.float32**])"

jmunetong: Thank you! I have modified the test to include your feedback.

airMeng (approved):

PR #224, opened 2026-05-28
 Add fused store_cache_xpu kernel

jmunetong:
## Summary

Add a SYCL kernel that fuses the two separate `aten::_index_put_impl_` calls (K write + V write) into a single kernel launch for KV-cache writes on XPU.

## Changes

- `src/sycl/KVCache.cpp`: new `StoreCacheKernel` + `store_cache_xpu` entry point (~75 lines)
- `include/sgl_kernel_ops.h`: forward declaration
- `src/torch_extension_sycl.cc`: op registration under `sgl_kernel` TORCH_LIBRARY_FRAGMENT
- `python/sgl_kernel/elementwise.py`: Python wrapper with docstring
- `python/sgl_kernel/__init__.py`: export

## Kernel details

- One work-group per token, work-items stride over `row_dim`
- Handles negative indices (skip token for speculative decoding)
- Supports bf16 and fp16 via `SYCL_DISPATCH_FLOATING_TYPES`
- Mirrors the CUDA `store_cache` JIT kernel from `sglang/jit_kernel/kvcache.py`

## Companion sglang PR

- [sglang #26594](https://github.com/sgl-project/sglang/pull/26594): adds `_is_xpu` to the `store_cache` dispatch gate in `memory_pool.py`

## Test

- Bit-parity verified: `store_cache_xpu` output matches `k_cache[indices] = k; v_cache[indices] = v` reference
- E2B GSM8K 5-shot (N=200): 0.175 accuracy — no regression

PR #236, opened 2026-06-04
store_cache: support strided (non-contiguous) K/V

jmunetong:
Address source rows by their real row stride instead of assuming row_dim, so per-head slices (e.g. Gemma SWA, stride=(2560,1)) work without a copy. Require only inner-row contiguity; gate the 16-byte OWord fast path on alignment with a scalar fallback. Add strided tests. 
Below are the performance results for both strided and non-strided conditions under the against 2xinput (baseline) kernels.

### Strided — SWA per-head slice (fast path, stride = heads × 512, 16B-aligned), row_dim = 512

| num_tokens | index_put_x2 µs | copy + fused µs | **fused_1launch µs** | **speedup** | vs copy+fused | fused GB/s |
|---:|---:|---:|---:|---:|---:|---:|
| 1    | 11.70  | 5.69   | **5.75**  | **2.04** | 0.99× | 0.7   |
| 8    | 12.82  | 8.87   | **5.88**  | **2.18** | 1.51× | 5.6   |
| 64   | 17.16  | 9.46   | **6.63**  | **2.59** | 1.43× | 39.5  |
| 512  | 41.60  | 19.81  | **14.01** | **2.97** | 1.41× | 149.6 |
| 4096 | 224.30 | 104.73 | **70.10** | **3.20** | 1.49× | 239.4 |

### Findings

- **Contiguous ≈ strided at the same width.** At row_dim = 512 the strided kernel tracks the
  contiguous kernel within noise (4096 tok: 70.1 vs 68.9 µs, 239 vs 243 GB/s) — real-stride
  addressing on the aligned OWord path is essentially free, reconfirming PR #236 adds no
  contiguous regression.
- **Speedup scales 2.0× → 3.25× with batch** in both modes: launch-overhead-bound at
  `num_tokens=1` (`fused_1launch` holds flat at ~5.8 µs while `index_put_x2` grows),
  bandwidth-bound by 4096.
- **Strided removes the copy tax:** `fused_1launch` beats `copy + fused` (`.contiguous()` + kernel)
  by **~1.4–1.5×** at every batch > 1 — the penalty a contiguity-requiring kernel (#224) would pay
  on SWA layers.


airMeng (commented):
Any performance differences before and after?

jmunetong:
@airMeng I have updated the PR description with a few results. In both cases, we see store_cache provides an improvement. We will have a decrease in performance for non-contiguous vectors (specifically in SWA layers as in gemma-4) but this condition is still better than two independent launches.

airMeng (approved):

PR #248, opened 2026-06-22
Add fast_topk SYCL support

cyxlily:
Adds XPU (SYCL) support for the fast_topk operator family used by DeepSeek-V3.2 sparse attention, matching the CUDA implementation in sglang/sgl-kernel/csrc/elementwise/topk.cu.

Note: we found a hardware/compiler ordering bug: after item.barrier(local_space), 32 subgroups issuing concurrent SIMD1 SLM loads to the same address occasionally observe the pre-workgroup HW-default 0 instead of the value written before the barrier. Traced by https://github.com/intel/llvm/issues/22558
Workaround: only let subgroup 0 perform one SIMD1 load and save topk to SLM, then all other subgroups read topk from SLM, ensuring topk is updated correctly.


tests/test_fast_topk_bench.py
```diff
@@ -0,0 +1,359 @@
...
+    )
+
+
+# -----------------------------------------------------------------------------
+# Generic XPU/CUDA event-based timer
+# -----------------------------------------------------------------------------
+def _sync():
+    if device.type == "xpu":
+        torch.xpu.synchronize()
+    elif device.type == "cuda":
```
jianan-gu: no need for cuda type, test for xpu only

src/sycl/TopKTransform.cpp
jianan-gu, line 1: Can we rename this file and also frontend, `topk ` is too common

jianan-gu (approved):

PR #256, opened 2026-06-26
ci: harden fused_moe benchmark flake (empty_cache + pipefail + missing &&)

arathi-hlab:
## Summary

The XPU CI flakes whenever `bench_fused_moe.py` OOMs mid-sweep on the shared `bmg-754` runner. The Python crash is hidden by `| tee`, then `update_baseline_from_log.py` raises on the truncated log in a downstream step. Three independent bugs compound — fixed together here.

1. **`benchmark/bench_fused_moe.py`** — `torch.xpu.empty_cache()` at the top of each config. Carry-over allocator fragmentation on a 12 GiB BMG is what causes the OOM; releasing cached blocks between configs gives the next one the full free pool.
2. **`.github/workflows/pr-test-xpu.yml` — `set -o pipefail`** — without it, `python3 ... | tee` returns `tee`'s exit code, so a Python crash silently passes the bench step.
3. **`.github/workflows/pr-test-xpu.yml` — 17 missing `&& \` joiners** — from `bench_moe_sum_reduce.py` onward, lines were joined with only `\`, folding the remaining 17 benchmarks into one `tee` argument list (dead code). All 24 benchmarks now actually run.

## Test plan
- [ ] CI runs the benchmark step end-to-end with all 24 scripts
- [ ] `bench_fused_moe.py` sweep completes without OOM, or fails loudly at the crash point
- [ ] `update_baseline_from_log.py` produces the same baseline comparison as today

### DIFF
--- a/.github/workflows/pr-test-xpu.yml
+++ b/.github/workflows/pr-test-xpu.yml
@@ -57,6 +57,7 @@ jobs:
         run: |
           docker exec -w /root/sglang ci_sglkernel_xpu \
             /bin/bash -c "\
+              set -o pipefail && \
               /miniforge3/envs/py3.10/bin/python3 -m pip install tabulate && \
               cd /root/sglang/sgl-kernel-xpu/benchmark && \
               python3 bench_flash_attn.py 2>&1 | tee flash.log && \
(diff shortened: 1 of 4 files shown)


sunjiweiswift (approved):

sunjiweiswift:
@claude  push empty commit for trigger CI

PR #266, opened 2026-07-06
default ApplyShuffleMulSum weight to 1.0 when factors is null

jiayisunx:
Set the default ApplyShuffleMulSum weight to 1.0 when factors is null, to align with the CUDA implementation.

### DIFF
--- a/src/sycl/MoEPrepareInputs.cpp
+++ b/src/sycl/MoEPrepareInputs.cpp
@@ -619,7 +619,7 @@ struct ApplyShuffleMulSum {
     float weights[MAX_TOPK];
     for (int k = 0; k < topk_ && k < MAX_TOPK; ++k) {
       src_indices[k] = static_cast<int>(dst2src_map_[out_tkn_id * topk_ + k]);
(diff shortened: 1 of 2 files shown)


kareemshaik80 (approved):

src/sycl/MoEPrepareInputs.cpp
```diff
@@ -619,7 +619,7 @@ struct ApplyShuffleMulSum {
     float weights[MAX_TOPK];
     for (int k = 0; k < topk_ && k < MAX_TOPK; ++k) {
       src_indices[k] = static_cast<int>(dst2src_map_[out_tkn_id * topk_ + k]);
-      weights[k] = (factors_ != nullptr) ? static_cast<float>(factors_[out_tkn_id * topk_ + k]) : 0.0f;
+      weights[k] = (factors_ != nullptr) ? static_cast<float>(factors_[out_tkn_id * topk_ + k]) : 1.0f;
```
mingfeima, line 622: this looks like a bug, what it is not exposed in test cases?
check if we need to update test case.

jiayisunx: I have updated UT to cover the null-factors case, and the CI is still having issues, and the failures don't seem to be related to my PR.

mingfeima (approved):

PR #267, opened 2026-07-06
[XPU] Enable MXFP4 GroupGemm ActType=2 (swiglu_gpt_oss) + WithBias for gpt-oss

libinta:
The MXFP4 W4A16 grouped-GEMM AOT instantiation matrix was pruned to {ActType 0 (silu), 4 (swiglu_deepseek_v4)} x {WithBias false} to keep Level Zero per-context module count in budget under TP>1.

gpt-oss-20b (GptOssForCausalLM) cannot run on this pruned kernel: its MoE uses the gpt-oss gated activation (swiglu_gpt_oss, ActType=2) with per-channel mlp1/mlp2 biases. sglang's mxfp4 path forwards the model's gemm1_alpha/gemm1_limit to sgl_kernel.fused_experts, which resolves to activation_type=2 + with_bias=true and dispatches
moe_grouped_mm_nt_xe20_mxfp4_w4a16. Against the pruned build this aborts at the first MoE forward:

  RuntimeError: mxfp4 fused kernel built with ActType=0 (silu) and
  ActType=4 (swiglu_deepseek_v4) only; got ActType=2.

(verified empirically: built pruned HEAD, launched gpt-oss-20b on Intel BMG/Xe2, observed the abort; re-enabling act=2+bias makes it serve.)

This re-adds ActType=2 and WithBias={false,true} to the cmake instantiation matrix and relaxes the matching dispatch-side TORCH_CHECKs. At TP=1 the larger module set fits; if TP>1 hits L0 module-pool exhaustion, narrow the matrix back down (see the cmake header comment).

### DIFF
--- a/src/GroupGemmMxfp4W4A16Xe20.cmake
+++ b/src/GroupGemmMxfp4W4A16Xe20.cmake
@@ -21,24 +21,28 @@ endfunction()
 #   <_256, _64,  _32> / SG_8_2_1   — avg_m > 128,  fuse_act=true
 #   <_256, _256, _32> / SG_8_4_1   — avg_m > 128,  fuse_act=false
 
-# TEMPORARY L0-module-pressure workaround: generate only the variants DSV4
-# actually dispatches (activation_type=0 silu and activation_type=4
-# swiglu_deepseek_v4, with_bias=false), skipping the full act × bias cartesian
-# product. This keeps the AOT-compiled MXFP4 .so count small, which is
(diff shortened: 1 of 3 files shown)


airMeng (commented):
Could you add related UT to monitor this feature

libinta:
> Could you add related UT to monitor this feature

@airMeng thanks for the review, the UT is added

airMeng (approved):

PR #276, opened 2026-07-08
per_token_quant_fp8: add XPU SYCL kernel

sspintel:
## Summary

Implements the SYCL kernel for `sgl_per_token_quant_fp8` on Intel XPU (Xe2/BMG). The op was previously a stub — declared in `sgl_kernel_ops.h`, wrapped in `gemm.py`, and exported, with a test + benchmark present, but **no SYCL implementation and no op registration**. This PR fills in the kernel, registers it, and adapts the test/benchmark to run on XPU.

Port of `sgl-kernel` `csrc/gemm/per_token_quant_fp8.cu`. Per-token (per-row) dynamic FP8 E4M3 quantization:

```
scale = rowmax(|x|) / 448
q[i]  = clamp(x[i] / scale, -448, 448)  as e4m3
```

## Details

- **`src/sycl/per_token_quant_fp8.cpp`** (new) — two kernels mirroring the CUDA reference:
  - warp kernel: one sub-group (32 lanes) per token, `TOKENS_PER_WG=8` — for large batches
  - block kernel: one work-group per token with block reduce — for small batches
  - device-preferred vector width with divisibility fallback; absmax reductions in fp32
- **`src/torch_extension_sycl.cc`** — `m.def`/`m.impl` registration with `Tensor(a!)`/`Tensor(b!)` mutable-output annotations
- **`tests/test_per_token_quant_fp8.py`** — fixed a `device = device` `UnboundLocalError`; parametrized over fp16 + bf16 and added an 8192-token case so both kernel paths are exercised
- **`benchmark/bench_per_token_quant_fp8.py`** — XPU event-based timing + pure-torch correctness check (dropped the CUDA-only vllm dependency)

## FP8 conversion

Investigated the SYCL-TLA `quantize` API and native SYCL fp8. DPC++ 2025.3 has no native fp8, and SYCL-TLA's quantize path is coupled to GEMM `SubgroupTensor`/TV-layout machinery (its E4M3 atoms are inline-asm epilogue fragments, not standalone). Used the CUTLASS `float_e4m3_t` cast that the rest of the repo already relies on (`per_tensor_quant_fp8.cpp`, group-quant kernels) — numerically validated against the torch reference.

## Performance (Intel Arc B580, ~210 GB/s DRAM copy ceiling)

Two-pass global design runs at **~85% of the copy ceiling** in the bandwidth-bound large-batch regime (e.g. 32768×4096 ≈ 143 GB/s counting read+write+scale, higher counting the L2-served reload). An SLM-cached single-pass variant was measured and rejected — per-token SLM footprint collapses work-group occupancy and is net slower. Documented in a code comment.

## Test plan

- [x] `tests/test_per_token_quant_fp8.py` — 40/40 pass on Intel Arc B580 (fp16 + bf16 × {128,256,512,8192} tokens × {512,1076,1368,2048,4096} hidden)
- [x] `benchmark/bench_per_token_quant_fp8.py` runs cleanly with correctness check
- [x] `pre-commit run --all-files` clean


src/sycl/per_token_quant_fp8.cpp
```diff
@@ -0,0 +1,312 @@
...
+        break;                            \
+      default:                            \
+        if (use_warp_kernel)              \
+          LAUNCH_WARP(T, DST_DTYPE, 1);   \
+        else                              \
+          LAUNCH_BLOCK(T, DST_DTYPE, 1);  \
+        break;                            \
+    }                                     \
+  } while (0)
+
```
pralay-das: The LAUNCH_FOR_VEC macro repeats the if (use_warp_kernel) branch 5×. can we use this kind of pattern ?
```cpp
#define LAUNCH_SWITCH(LAUNCH, T, DST_DTYPE)        \
  switch (vec_size) {                              \
    case 16: LAUNCH(T, DST_DTYPE, 16); break;      \
    case 8:  LAUNCH(T, DST_DTYPE, 8);  break;      \
    case 4:  LAUNCH(T, DST_DTYPE, 4);  break;      \
    case 2:  LAUNCH(T, DST_DTYPE, 2);  break;      \
    default: LAUNCH(T, DST_DTYPE, 1);  break;      \
  }

#define LAUNCH_FOR_VEC(T, DST_DTYPE)               \
  do {                                             \
    if (use_warp_kernel)                           \
      LAUNCH_SWITCH(LAUNCH_WARP, T, DST_DTYPE);    \
    else                                           \
      LAUNCH_SWITCH(LAUNCH_BLOCK, T, DST_DTYPE);   \
  } while (0)
```

tests/test_per_token_quant_fp8.py
```diff
@@ -37,16 +37,21 @@ def sglang_per_token_quant_fp8(
...
+    list(itertools.product([128, 256, 512, 8192], [512, 1076, 1368, 2048, 4096])),
 )
 def test_per_token_quant_compare_implementations(
+    dtype: torch.dtype,
     num_tokens: int,
     hidden_dim: int,
 ):
-    device = device
-    x = torch.rand((num_tokens, hidden_dim), dtype=torch.float16, device=device)
+    x = torch.rand((num_tokens, hidden_dim), dtype=dtype, device=device)
```
pralay-das, line 55: as we are creating random tensor , can we add `torch.manual_seed(42)` to make it deterministic?

pralay-das:
also add `pytest` file and `benchmark` file in `run_suite.py` and `pr-test-xpu.yml`

pralay-das (approved):

PR #214, opened 2026-05-21
Add fp8_mqa_logits kernels for GLM NSA

Xia-Weiwen:
Add fp8_mqa_logits kernels for GLM5.1 native sparse attention
- `fp8_mqa_logits` for prefill (python only implementation with torch._scaled_mm)
- `fp8_paged_mqa_logits` for decode (SYCL-TLA implementation as torch._scaled_bmm is not available yet)

Test: tests/test_fp8_mqa_logits.py
Benchmark: benchmark/bench_fp8_mqa_logits.py

PR #279, opened 2026-07-09
Improve fp8_mqa_logits by handling OOM and tensor dimension mismatch

Xia-Weiwen:
As the title.


tests/test_fp8_mqa_logits.py
polisettyvarma, line 209: can we parameterize ?

python/sgl_kernel/nsa.py
polisettyvarma, line 145: can we handle OOM here also ?

mingfeima (approved):
the whole thing is a WA upon a naive impl. We can do this for now. Put a TODO on the top of this file, we need to provide native kernels for it in future.

this nsa is not going to be performant.

Xia-Weiwen:
> the whole thing is a WA upon a naive impl. We can do this for now. Put a TODO on the top of this file, we need to provide native kernels for it in future.
> 
> this nsa is not going to be performant.

Thanks for reviewing. According to the discussion with @airMeng in https://github.com/sgl-project/sgl-kernel-xpu/pull/214, the performance of `torch._scaled_mm` is good enough. I had a SYCL-TLA-based implementation at the beginning, but it turned out to be less performant than the oneDNN-based `torch._scaled_mm`. So, we decided to use a python/PyTorch implementation for prefill (`fp8_mqa_logits`) for both simplicity and performance. For decode (`fp8_paged_mqa_logits`), we will evaluate `torch._scaled_bmm` in the future when it is available in oneDNN and PyTorch XPU team (they are WIP).

mingfeima:
In future, we need to have the capability to implement customized kernels with better perf than oneDNN. otherwise ...

python/sgl_kernel/nsa.py
Xia-Weiwen: Thanks for the comment. The OOM pressure is much weaker here. OOM only happens when batch size is really big. In that case, OOM is expected. And in that case, OOM can probably happen elsewhere rather than here.

tests/test_fp8_mqa_logits.py
Xia-Weiwen: Thanks. Updated.

Xia-Weiwen:
Hi @polisettyvarma The CI passed. I will land this now. For any further improvements or bug fixes, I will have new PRs for them.

PR #280, opened 2026-07-09
DeepSeek V4 Add fused_q_indexer_rope_hadamard_quant support

polisettyvarma:
Add unit test and run in CI


python/sgl_kernel/fused_q_indexer_rope_hadamard_quant_torch.py
airMeng, line 1: this is a native pytorch implementation?

polisettyvarma: yes until SYCL kernel is implemented. we want DS V4 to enable functionally first in upstream.

mkumargarg (approved):

PR #281, opened 2026-07-10
Improve small k perf for group gemm

gaopengff:
This is from https://github.com/vllm-project/vllm-xpu-kernels/pull/340 and it looks like a cutlass issue.
For now, **XE_STORE_2D<16, 8, 32>** has better performance than **make_block_2d_copy_D** when k of gemm is small.
For qwen-3-30b-a3b with tp=4 and K=192, while setting bs=12 and input length to 4096, the M,N,K would be(49152, 2048, 192). The perf improved from 9.2ms to 6.1ms, which affects TTFT.

We can remove this if sycl-tla fix the bug of make_block_2d_copy_D.

### DIFF
--- a/src/sycl/kernels/moe/xe20/bf16/moe_kernel.hpp
+++ b/src/sycl/kernels/moe/xe20/bf16/moe_kernel.hpp
@@ -31,6 +31,7 @@
  **************************************************************************************************/
 #pragma once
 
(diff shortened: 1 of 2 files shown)


airMeng:
replaced by https://github.com/sgl-project/sgl-kernel-xpu/pull/365

PR #284, opened 2026-07-11
Add qk_norm_rope support for Flux

CaoE:
### Motivation
Add `fused_qk_norm_rope_with_cos_sin_cache_inplace` to fuse QK RMSNorm + RoPE with a precomputed cos/sin cache for the FLUX model.
Note: There is a legacy SYCL Kernel for Fused QK Norm + RoPE which uses packed QKV and YaRN RoPE frequencies computed on the fly, rather than read from a cache. To maintain code readability, no implementation reuse is enforced. Furthermore, considering that the current SGLANG does not use this legacy SYCL kernel, it will be retained for now and may be removed in the future.

Generated by Agent.

### Test
Added `test_fused_qk_norm_rope_with_cache`, `test_fused_qk_norm_rope_with_cache_last_dim_strided` and benchmarks.
Overall average speedup: 5.74x
Overall max speedup:     22.32x
Overall min speedup:     3.05x

### DIFF
--- a/benchmark/bench_fused_qk_norm_rope.py
+++ b/benchmark/bench_fused_qk_norm_rope.py
@@ -2,9 +2,10 @@ import itertools
 import os
 
 import pandas as pd
+import sgl_kernel
 import torch
 import triton
-from sgl_kernel import fused_qk_norm_rope
+from bench_fused_qk_rope_with_cache import create_cos_sin_cache
 
 # Supported dtypes and their properties
 DTYPE_MAP = {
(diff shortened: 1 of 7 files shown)


src/sycl/FusedQKNormRope.cpp
```diff
@@ -291,84 +539,388 @@ void fused_qk_norm_rope(
...
+  TORCH_CHECK(q_weight.size(0) == head_dim, "q_weight size must match head_dim");
+  TORCH_CHECK(k_weight.size(0) == head_dim, "k_weight size must match head_dim");
+  TORCH_CHECK(cos_sin_cache.dim() == 2, "cos_sin_cache must be 2D [max_position, rope_dim]");
+  const int64_t rope_dim = cos_sin_cache.size(1);
+  TORCH_CHECK(rope_dim % 2 == 0, "rope_dim must be even");
+  TORCH_CHECK(rope_dim <= head_dim, "rope_dim must be <= head_dim");
+  TORCH_CHECK(positions.dim() == 1, "positions must be 1D [num_tokens]");
+  TORCH_CHECK(positions.size(0) == num_tokens, "positions size must match flattened q/k tokens");
+
+  check_subgroup_size_supported(NUM_REDUCE_STAGES);
```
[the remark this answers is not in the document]
CaoE: This should be guaranteed by the caller; otherwise, checking it here would impact performance.

src/sycl/FusedQKNormRope.cpp
```diff
@@ -291,84 +539,388 @@ void fused_qk_norm_rope(
...
+    torch::Tensor& positions,
+    bool is_neox,
+    double eps) {
+  TORCH_CHECK(q.dim() == k.dim(), "q and k must have the same rank, got q:", q.dim(), " k:", k.dim());
+  TORCH_CHECK(q.dim() == 3 || q.dim() == 4, "q/k must be 3D or 4D tensors, got q:", q.dim());
+  TORCH_CHECK(q.scalar_type() == k.scalar_type(), "q and k must have the same dtype");
+  TORCH_CHECK(q_weight.scalar_type() == q.scalar_type(), "q_weight dtype must match q dtype");
+  TORCH_CHECK(k_weight.scalar_type() == k.scalar_type(), "k_weight dtype must match k dtype");
+  TORCH_CHECK(cos_sin_cache.scalar_type() == at::ScalarType::Float, "cos_sin_cache must be float32");
+
```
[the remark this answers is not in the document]
CaoE: In many cases, cos_sin_cache is in fp32, so fp32 is asserted here. More dtypes for cos_sin_cache can be supported in the future if needed.

ck-intel (approved):
LGTM, verified the changes.

benchmark/bench_fused_qk_norm_rope.py
```diff
@@ -300,6 +301,216 @@ def calculate_effective_bandwidth(
     }
 
 
+def apply_rotary_emb_from_cache(
+    x: torch.Tensor,
+    cos_sin_cache: torch.Tensor,
```
sunjiweiswift, line 306: which dtype here? 
In CUDA, this part always uses FP32; we should also ensure that the accumulation is performed in FP32.

sunjiweiswift (approved):

benchmark/bench_fused_qk_norm_rope.py
CaoE: The dtype of cos_sin_cache is fp32, and generated by `create_cos_sin_cache`.

benchmark/bench_fused_qk_norm_rope.py
```diff
@@ -300,6 +301,216 @@ def calculate_effective_bandwidth(
...
+    positions: torch.Tensor,
+    rope_dim: int,
+    is_neox_style: bool,
+) -> torch.Tensor:
+    half_rope_dim = rope_dim // 2
+    pos = positions.to(device=x.device, dtype=torch.long).view(-1)
+    cos = cos_sin_cache[pos, :half_rope_dim].to(dtype=x.dtype)
+    sin = cos_sin_cache[pos, half_rope_dim:].to(dtype=x.dtype)
```
sunjiweiswift, line 314: also here.  @gaopengff  pls review

sunjiweiswift:
I've added the `run_ci` label
you can trigger it by push a empty commit.
@CaoE

benchmark/bench_fused_qk_norm_rope.py
CaoE: This is for benchmark.

PR #286, opened 2026-07-14
[Attn] optm for non paged attn

Valentine233:
Add fast paths for pure-decode and pure-prefill, to avoid redundant sub-kernels launch overhead.
Note that this also benefits the paged case, where the overhead part is relatively small.
Optimize Flux.2 case https://jira.devtools.intel.com/browse/SGLANGT-1383, from 37.03 to 62.98 TFLOP/s.

Non paged performance:

- decode: medium speedup 1.52x, maximum speedup 5.43x;
- prefill: medium speedup 1.07x, maximum speedup 2.15x.

### DIFF
--- a/src/sycl/flash_attention.cpp
+++ b/src/sycl/flash_attention.cpp
@@ -1233,9 +1233,7 @@ std::vector<at::Tensor> mha_fwd(
     int const sm_margin,
     std::optional<at::Tensor> out_ = std::nullopt) {
   // Supports both paged (page_table != None) and non-paged (contiguous ragged
-  // KV, page_table == None) layouts. In the non-paged case both sub-launches run
-  // the non-paged prefill kernel (decode is paged-only), which is mathematically
-  // identical for the seqlen_q <= 1 decode batches.
+  // KV, page_table == None) layouts.
   // ``seqlens_rotary_`` is intentionally not checked here: callers pass it
   // alongside ``cache_seqlens`` even when rotary is disabled, and the
   // sub-kernels only consume it inside the ``rotary_cos_.has_value()`` branch.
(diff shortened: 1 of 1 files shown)


src/sycl/flash_attention.cpp
```diff
@@ -1293,6 +1291,33 @@ std::vector<at::Tensor> mha_fwd(
         std::move(skip_mask));
   };
 
+  // Pure-decode fast path: max_seqlen_q is the MAX seqlen_q over all batches, so
+  // max_seqlen_q <= 1 guarantees EVERY batch is decode (no seqlen_q > 1 prefill
+  // rows) for any batch_size. The prefill sub-launch below would then launch a
+  // full kernel grid only to `continue` past every tile — pure overhead. Run
+  // just the decode kernel. Gate uses only the host scalar max_seqlen_q, so it
```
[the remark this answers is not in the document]
Valentine233: Fixed.

ck-intel (approved):
LGTM, performance verified on Flux.2-dev model.

src/sycl/flash_attention.cpp
sunjiweiswift: Can the conditional check for `page` be removed here, rather than checking for `decode` within `chunkprefill`?

sunjiweiswift: If chunk-prefill is not required when `no_page` , the following logic applies:
if (q == 1) // both page and no_page Or separately
decode::mha_fwd
if (!page)
prefill::mha_fwd
else (page)
chunkprefill::mha_fwd

sunjiweiswift (approved):

src/sycl/flash_attention.cpp
Valentine233: Thanks, updated.

Valentine233: Thanks, updated.

PR #287, opened 2026-07-14
[ROPE] cos_sin_cache dtype bf16-> fp32

sunjiweiswift:

PR #290, opened 2026-07-15
Use fp32 dtype for cos_sin_cache  in rope

gaopengff:


gaopengff:
Close this as https://github.com/sgl-project/sgl-kernel-xpu/pull/287 has done the same work.

PR #292, opened 2026-07-15
[Internal/CRI] Changes to build/enable CRI - WIP

mkumargarg:
These are the changes required to build/enable for cri.


airMeng:
I'd suggest hold on CRI upstream since there is still long time before we have CRI machines for upstream CI/CD

mkumargarg:
raised the new one with only the build changes. https://github.com/sgl-project/sgl-kernel-xpu/pull/390

PR #311, opened 2026-07-18
[XPU] Add fused MXFP8 W8A16 GEMM (moe_grouped_mm_nt_xe20_mxfp8_w8a16)

kalyank007:
[XPU] Add fused MXFP8 W8A16 GEMM (moe_grouped_mm_nt_xe20_mxfp8_w8a16)

Fused bf16-activation x E4M3-weight GEMM with per-32-K-block fp32
direct-multiplier scales, dequantizing B to bf16 in registers before the
DPAS MMA. Fork of the MXFP4 W4A16 grouped GEMM: the B fragment element
type is cutlass::float_e4m3_t (one byte/elem, row stride K, no nibble
unpack), relying on Xe_Reorder<float_e4m3_t, bfloat16_t> for the
in-register upcast. The dense linear path calls it with n_experts=1.

- src/sycl/kernels/moe/xe20/mxfp8_w8a16/{moe_mainloop,moe_kernel}.hpp:
  forked mxfp8 kernel + mainloop.
- src/sycl/GroupGemmMxfp8W8A16Xe20.cpp + LauncherInstance.cpp.in +
  src/GroupGemmMxfp8W8A16Xe20.cmake: dispatcher, launcher template, and
  tile instantiation (non-fused/silu slice, 3 tiles x {bias,no-bias}).
- torch_extension_sycl.cc / sgl_kernel_ops.h: op def/impl + decl.
- python/sgl_kernel/gemm.py: mxfp8_w8a16_gemm wrapper (2D [N,K] weight,
  fp32 [N,K/32] scales) wrapping the single-expert grouped call;
  exported from __init__.py.

Build against oneAPI 2025.3 (libsycl.so.8) to match the py312 torch-xpu
runtime. Validated vs the sglang torch dequant reference: cosine sim
1.000000 across decode/prefill/MLP shapes, and identical greedy output
to the Phase A path on full Llama-3.3-70B-Instruct (tp=8).

### DIFF
--- a/include/sgl_kernel_ops.h
+++ b/include/sgl_kernel_ops.h
@@ -388,6 +388,15 @@ void moe_grouped_mm_nt_xe20_mxfp4_w4a16(
     double gemm1_alpha = 1.702,
     double gemm1_limit = 7.0);
 
+void moe_grouped_mm_nt_xe20_mxfp8_w8a16(
+    torch::Tensor& output,
+    const torch::Tensor& activations,
+    const torch::Tensor& packed_weights,
+    const torch::Tensor& scales,
+    const std::optional<at::Tensor>& bias,
+    const torch::Tensor& total_rows_for_experts,
+    const int64_t n_experts);
+
 void prepare_moe_input(
     const torch::Tensor& topk_ids,
     torch::Tensor& expert_offsets,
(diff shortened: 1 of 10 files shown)


airMeng (commented):
Hi @kalyank007  Thank you for your contribution! We are working on compressing binary size as much as possible for our first release. Since MX-FP8 is not our priority on BMG, could you reduce the configurations to the least, to avoid enlarge our final binary too much? I will suggest 1 config for prefill, 1 for decoding is enough.

PR #315, opened 2026-07-21
Add fast path for bs=1 prefill in chunkprefill kernel

liangan1:
The benchmark for bs=1 prefills show that the additional decode kernel involves obvious overhead.  Let us take the bs=1, num_qkv_head=32, head_dim=128 as examples, the decoder kernel introduce about 45% overhead. 

I use the unitrace to collect the kernel time hw counter. The overhead is not purely from the host kernel launch. It issues a lot of intrinsic and cause obvious EU stalls. 

ALU1： 62.7M execution slots
ALU0：13.6M
SEND：6.7M

### DIFF
--- a/src/sycl/flash_attention.cpp
+++ b/src/sycl/flash_attention.cpp
@@ -1251,9 +1251,6 @@ std::vector<at::Tensor> mha_fwd(
   int64_t batch_size = cu_seqlens_q.size(0) - 1;
   TORCH_CHECK(batch_size >= 0, "cu_seqlens_q must have at least 1 element.");
 
-  auto seqlens_q = cu_seqlens_q.slice(0, 1, batch_size + 1).sub(cu_seqlens_q.slice(0, 0, batch_size));
-  auto is_prefill = seqlens_q.gt(1).contiguous();  // true for prefill batches
-
   // Forward every shared argument to a sub-kernel, overriding only the output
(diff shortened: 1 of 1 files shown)


src/sycl/flash_attention.cpp
```diff
@@ -1291,6 +1287,15 @@ std::vector<at::Tensor> mha_fwd(
         std::move(skip_mask));
   };
 
+  if (batch_size == 1 && max_seqlen_q > 1) {
+    auto out = launch(prefill::mha_fwd, std::move(out_), std::nullopt)[0];
+    auto empty_f = at::empty({0}, q.options().dtype(at::kFloat));
+    return {out, empty_f, empty_f, empty_f};
+  }
```
mingfeima, line 1294: Top-level `mha_fwd` has a fast path when `bs == 1 && max_seqlen_q > 1` #286, so `chunkprefill::mha_fwd` will not be used:

something like:

```cpp
if (max_seqlen_q == 1) {
  return dispatch(decode::mha_fwd, std::nullopt);
} else if (!page_table.has_value() || batch_size == 1) {
  return dispatch(prefill::mha_fwd, std::nullopt);
} else {
  return dispatch(chunkprefill::mha_fwd);
}

@Valentine233 comment?

liangan1: Good. My code base should be older than #286. let me close this pr.

PR #316, opened 2026-07-21
[MLA] Optimize sparse MLA

Valentine233:
### Description
Currently, sgl-kernel-xpu only has the sparse MLA decode with low perf.
This PR removes the old sparse MLA decode, and ports sparse MLA for both prefill and decode from XAttention.

### Performance
**Decode**
New kernel gets a speedup of **2.4x–10.6x**.
| batch | topk | extra_topk | new (ms) | old (ms) | new (GB/s) | old (GB/s) | speedup |
|------:|-----:|-----------:|---------------:|--------------:|-----------------:|----------------:|--------:|
| 1   | 64  | 0   | 0.0570 | 0.1393 | 1.24  | 0.51  | 2.44x |
| 1   | 64  | 512 | 0.0719 | 0.5750 | 5.17  | 0.65  | 8.00x |
| 1   | 128 | 0   | 0.0586 | 0.1408 | 1.84  | 0.77  | 2.40x |
| 1   | 128 | 512 | 0.0733 | 0.5783 | 5.58  | 0.71  | 7.89x |
| 1   | 512 | 0   | 0.0701 | 0.4598 | 4.76  | 0.73  | 6.56x |
| 1   | 512 | 512 | 0.0851 | 0.9030 | 7.46  | 0.70  | 10.61x |
| 8   | 64  | 0   | 0.0568 | 0.1398 | 9.92  | 4.03  | 2.46x |
| 8   | 64  | 512 | 0.0757 | 0.5817 | 39.28 | 5.11  | 7.68x |
| 8   | 128 | 0   | 0.0588 | 0.1459 | 14.70 | 5.93  | 2.48x |
| 8   | 128 | 512 | 0.0787 | 0.5874 | 41.58 | 5.57  | 7.46x |
| 8   | 512 | 0   | 0.0732 | 0.4719 | 36.48 | 5.66  | 6.45x |
| 8   | 512 | 512 | 0.0980 | 0.9206 | 51.81 | 5.52  | 9.39x |
| 42  | 64  | 0   | 0.0765 | 0.4157 | 38.66 | 7.12  | 5.43x |
| 42  | 64  | 512 | 0.3204 | 1.5799 | 48.70 | 9.88  | 4.93x |
| 42  | 128 | 0   | 0.0873 | 0.4385 | 52.03 | 10.35 | 5.02x |
| 42  | 128 | 512 | 0.3458 | 1.5959 | 49.69 | 10.77 | 4.62x |
| 42  | 512 | 0   | 0.2897 | 1.3007 | 48.40 | 10.78 | 4.49x |
| 42  | 512 | 512 | 0.5597 | 2.4871 | 47.65 | 10.72 | 4.44x |
| 128 | 64  | 0   | 0.1359 | 1.0897 | 66.38 | 8.28  | 8.02x |
| 128 | 64  | 512 | 1.1702 | 4.0080 | 40.64 | 11.86 | 3.42x |
| 128 | 128 | 0   | 0.2346 | 1.1352 | 58.99 | 12.19 | 4.84x |
| 128 | 128 | 512 | 1.2914 | 4.0542 | 40.56 | 12.92 | 3.14x |
| 128 | 512 | 0   | 1.0306 | 3.3342 | 41.47 | 12.82 | 3.24x |
| 128 | 512 | 512 | 2.0044 | 6.2775 | 40.55 | 12.95 | 3.13x |
| 256 | 64  | 0   | 0.3244 | 1.9629 | 55.60 | 9.19  | 6.05x |
| 256 | 64  | 512 | 2.3316 | 7.2853 | 40.79 | 13.06 | 3.13x |
| 256 | 128 | 0   | 0.6063 | 2.0593 | 45.64 | 13.44 | 3.40x |
| 256 | 128 | 512 | 2.5463 | 7.4137 | 41.13 | 14.13 | 2.91x |
| 256 | 512 | 0   | 2.0177 | 6.1699 | 42.36 | 13.85 | 3.06x |

**Prefill**
New kernel reaches bandwidth of **42–72 GB/s**.
| s_q | H | topk | new (ms) | new (GB/s) |
|-----:|--:|-----:|---------------:|-----------------:|
| 512  | 16  | 2048 | 17.122 | 71.59 |
| 512  | 32  | 2048 | 18.281 | 68.03 |
| 512  | 128 | 2048 | 31.930 | 42.30 |
| 2048 | 16  | 512  | 19.844 | 64.46 |
| 2048 | 128 | 512  | 28.654 | 62.06 |


src/MlaSparseDecodeXe20.cmake
```diff
@@ -1,21 +1,31 @@
...
+            endif()
+            if(IS_FP8_QUERY)
+                set(_q_tag "qfp8")
+            else()
+                set(_q_tag "qbf16")
+            endif()
+            set(GENERATED_FILE
+                "${CMAKE_CURRENT_BINARY_DIR}/sycl/sparse_mla_decode_fp8_fwd_${_topk_tag}k${D_QK}_${_q_tag}.cpp")
+            configure_file(${MLA_SPARSE_DECODE_TEMPLATE} ${GENERATED_FILE} @ONLY)
+            list(APPEND device_cpp_common ${GENERATED_FILE})
```
airMeng: ```device_cpp_common``` is for general SYCL kernels, ```device_cpp_xe20``` is for Xe20 only, could you check all variables here and update them all?

airMeng (changes requested):

Xia-Weiwen:
Isn't the India team working on it already?

src/MlaSparseDecodeXe20.cmake
airMeng: cc @pralay-das @jayachandranb-ai

Valentine233:
> Isn't the India team working on it already?

We decide to use the implementation from XAttention for better performance.

src/sycl/mla_sparse_prefill.cpp
```diff
@@ -0,0 +1,156 @@
...
+  at::Tensor lse = torch::empty({s_q, h_q}, opts.dtype(torch::kFloat));
+  at::Tensor max_logits = torch::empty({s_q, h_q}, opts.dtype(torch::kFloat));
+
+  // Chunk gathered_k along s_q to bound peak device memory.
+  // gathered_k is [chunk, topk, d_qk] in bf16; cap the workspace at ~256 MB.
+  constexpr int64_t PREFILL_GATHERED_K_MAX_BYTES = 256LL * 1024 * 1024;
+  const int64_t per_seq_gathered_bytes = static_cast<int64_t>(topk) * d_qk * 2;  // bf16 = 2 bytes
+  int chunk_size = std::max(1, static_cast<int>(PREFILL_GATHERED_K_MAX_BYTES / per_seq_gathered_bytes));
+  chunk_size = std::min(chunk_size, s_q);
+
```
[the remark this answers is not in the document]
Valentine233: Add assertion to ensure topk>0.

python/sgl_kernel/attention.py
```diff
@@ -333,24 +333,147 @@ def flash_mla_with_kvcache(
...
                 and extra_topk_length.shape == (B,)
             ), "extra_topk_length must be int32 on the same device with shape [B]"
-        torch.ops.sgl_kernel.flash_mla_sparse_decode.default(
-            out,
-            lse,
+        is_fp8_query = q.dtype == torch.float8_e4m3fn
+        out, lse_bshq = torch.ops.sgl_kernel.flash_mla_sparse_decode.default(
             q,
             k_cache,
             indices,
```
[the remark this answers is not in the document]
Valentine233: Modify to make sure q is bf16.

tests/run_suite.py
```diff
@@ -21,7 +21,6 @@ class TestFile:
         TestFile("test_flash_attn_sparse.py"),
         TestFile("test_flash_mla_decode.py"),
         TestFile("test_flash_mla_prefill.py"),
-        TestFile("test_flash_mla_with_kvcache.py"),
         TestFile("test_moe_align.py"),
```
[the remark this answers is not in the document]
Valentine233: Added the new sparse MLA tests.

pralay-das:
Hi @Valentine233, could you point it out how XAttention sparse algorithm is better than flashMLA sparse algorithm, is there any docs or reference?

Valentine233:
> Hi @Valentine233, could you point it out how XAttention sparse algorithm is better than flashMLA sparse algorithm, is there any docs or reference?

Hi @pralay-das , please refer to the perf result of sparse MLA decode in PR description.

pralay-das:
> > Hi @Valentine233, could you point it out how XAttention sparse algorithm is better than flashMLA sparse algorithm, is there any docs or reference?
> 
> Hi @pralay-das , please refer to the perf result of sparse MLA decode in PR description.

hi, I am really sorry, I am not asking about the how you have implemented it and getting better performance, I am more interested with how algorithm wise this implementation is better and where it is mentioned. [for example: from normal flash attention(SDPA) , flash attention2 is better and this paper(https://arxiv.org/pdf/[2205.14135](https://arxiv.org/pdf/2205.14135)) explained that], so wanted to know do you have any reference like that?

Valentine233:
> > > Hi @Valentine233, could you point it out how XAttention sparse algorithm is better than flashMLA sparse algorithm, is there any docs or reference?
> > 
> > 
> > Hi @pralay-das , please refer to the perf result of sparse MLA decode in PR description.
> 
> hi, I am really sorry, I am not asking about the how you have implemented it and getting better performance, I am more interested with how algorithm wise this implementation is better and where it is mentioned. [for example: from normal flash attention(SDPA) , flash attention2 is better and this paper(https://arxiv.org/pdf/[2205.14135](https://arxiv.org/pdf/2205.14135)) explained that], so wanted to know do you have any reference like that?

Thanks, got your point! Unfortunately, there is no such doc reference. As the previous perf result is low, we tried the XAttn kernel which proved good speedup. Actually, I haven't taken a deep look into the old kernels. If you are interested, please check the codes.

pralay-das:
Hi, thanks, a few observations. The perf numbers are compelling, but I'd hold this from merging in its current form. It replaces the established sycl-tla (CUTLASS-3.x collective-builder) structure — the same layout used by the non-sparse mla/, flash_attention_v2/, and the old mla_sparse/ — with a flat structure. That fork is what blocks generalizing to other Intel GPUs and creates correctness + maintainability debt. Details below.

**1. Architecture diverges from the repo's sycl-tla convention**

This PR deletes the entire collective/kernel/device layering for sparse MLA:

collective/xe_mla_sparse_mainloop.hpp, collective/xe_mla_sparse_epilogue.hpp
kernel/xe_mla_sparse_kernel.hpp, kernel/mla_sparse_tile_scheduler.hpp
device/mla_sparse_runner.hpp, device/mla_sparse_decode_types.hpp
and replaces them with two ~830-line flat headers (sparse_mla_decode_fp8_fwd.hpp, sparse_mla_prefill_fwd.hpp) that:

launch cutlass::device_kernel directly (no device/ runner, unlike mla_decode.cpp → MLA<Kernel>::run),
use a flat POD SparseAttnFwdParams copied verbatim from upstream (flash.h: "Same as deepseek-ai/FlashMLA/csrc/params.h#L145") instead of the CUTLASS Arguments→Params/to_underlying_arguments() split every other kernel uses,
live in a new flash_xpu namespace island (via namespace_config.h) instead of cutlass::flash_attention::*,
carry third-party CUDA lineage (flash_common_xpu.hpp: Copyright (c) 2024, Tri Dao; // leverage from .../FlashMLA/...), which the native sycl-tla path had deliberately eliminated.
The repo now has to maintain two attention idioms, with no shared collective/epilogue/scheduler and no reuse of the MLA runner's split-KV / workspace machinery.

**2. This structure doesn't generalize to other Intel GPUs**

mla/collective/xe_mla_mainloop.hpp dispatches on a DispatchPolicy (XeDefault<Stages>) template that is designed to be specialized per architecture — bringing up a new Intel GPU means adding a DispatchPolicy specialization, not editing the kernel body. The flat kernels here instead hardcode SYCL_INTEL_TARGET 20 and bake their MMA/copy atoms and tiling inline, so supporting another Intel GPU means forking and re-tuning the whole kernel rather than adding a policy specialization.

**3. The flash_mla_with_kvcache compat path is currently dead**

attention.py:305 asserts k_cache.dtype == torch.uint8, but the op it calls requires kv.dtype() == Float8_e4m3fn (mla_sparse_decode.cpp:55) — mutually exclusive, so no call through the wrapper can succeed, and no test covers it. This is the entry point downstream mla_decode users actually call, so it must round-trip and be tested before merge.

**4. Most importantly — the staged gather→HBM→attention design is a structural concern**

The gather kernel materializes the full dense gathered_k to global memory (bf16), and the attention kernel reads it back — a mandatory extra HBM round-trip of the KV working set. The deleted fused XeMlaSparseMainloop gathered directly into registers/SLM inside the attention loop. The dequant itself is necessary work; the avoidable cost is spilling the dequantized bf16 result to global at full 512×2 B/token width and re-reading it, versus dequantizing into registers/SLM inside the mainloop.

Memory scaling is a real, verifiable problem here:

Decode gathered_k is uncapped — mla_sparse_decode.cpp:120 allocates [b, s_q, topk+extra_topk, 512] bf16 in one shot (e.g. b=128, topk=512 → 67 MB; b=64, topk=2048+512 → 168 MB). It grows linearly with b·s_q·topk, so this will hit OOM at larger batch/topk. There's no chunking here at all.
Prefill had to add a 256 MB workspace cap with s_q chunking (mla_sparse_prefill.cpp:76) precisely because the intermediate is too big — which serializes the work: e.g. s_q=512, topk=2048 → chunk ≈ 113 rows → ⌈512/113⌉ = 5 sequential gather+attention launch pairs, each a full grid launch + barrier, adding launch overhead and killing cross-chunk overlap.
To be clear about what is and isn't measured: the HBM round-trip and the memory-scaling numbers above are facts from the code. Whether removing them actually wins is not something I've benchmarked — it's a hypothesis. But it's a well-founded one, and it exposes a gap in the PR's justification: the case for dropping sycl-tla is "XAttention is faster," which conflates two independent variables — the algorithm (XAttention vs. the old sparse math) and the execution structure (staged-to-HBM vs. fused). The measured speedup could be coming entirely from the algorithm while the staged structure leaves performance on the table. Before we accept a structural fork that costs us the collective-builder generality, I'd like to see a fused sycl-tla implementation of the XAttention algorithm as an apples-to-apples baseline — that's the experiment that would actually settle it.

pralay-das (changes requested):
change required.

Valentine233:
Thanks for the comment, @pralay-das .

I fixed the k_cache dtype bug and avoided decode OOM risk.
This PR is critical for now because it targets to optimize DeepSeekV4.
For the structure, I doubt that the existing sycl-tla one is the best one, and the config tuning is always required for other intel GPUs whatever the structure is. In addition, the new kernel decouples the irregular indexed gather + fp8 dequant from the regular dense attention, which could improve efficiency. Could you please give some inputs, @mingfeima ?

mkumargarg:
Hi, this PR is completely deleting the old code so it should not be merged. Also, in my opinion, fused version (gather with parallel to compute) should always be better in performance once optimized than this implementation, hence the old code should not be removed.
Moreover, please discuss on the points and concerns raised by @pralay-das and if required work with him to align the design and implementation so that our future optimization work also not get impacted. @pralay-das please discuss and work with @Valentine233 to incorporate some of the review comments which are must as of now and rest you can take care while further optimizing the code. Thanks.
cc: @Valentine233 @pralay-das @mingfeima

src/MlaSparseDecodeXe20.cmake
```diff
@@ -1,21 +1,31 @@
...
+            endif()
+            if(IS_FP8_QUERY)
+                set(_q_tag "qfp8")
+            else()
+                set(_q_tag "qbf16")
+            endif()
+            set(GENERATED_FILE
+                "${CMAKE_CURRENT_BINARY_DIR}/sycl/sparse_mla_decode_fp8_fwd_${_topk_tag}k${D_QK}_${_q_tag}.cpp")
+            configure_file(${MLA_SPARSE_DECODE_TEMPLATE} ${GENERATED_FILE} @ONLY)
+            list(APPEND device_cpp_xe20 ${GENERATED_FILE})
```
airMeng: Can you help to update all cmakes of other attention variants here?

Valentine233: Done.

airMeng (approved):

mingfeima:
> Hi, this PR is completely deleting the old code so it should not be merged. Also, in my opinion, fused version (gather with parallel to compute) should always be better in performance once optimized than this implementation, hence the old code should not be removed. Moreover, please discuss on the points and concerns raised by @pralay-das and if required work with him to align the design and implementation so that our future optimization work also not get impacted. @pralay-das please discuss and work with @Valentine233 to incorporate some of the review comments which are must as of now and rest you can take care while further optimizing the code. Thanks. cc: @Valentine233 @pralay-das @mingfeima

shall we move the discussion to Teams not on github?

src/sycl/kernels/mla_sparse/sparse_mla_prefill_fwd.hpp
```diff
@@ -0,0 +1,813 @@
...
+  compat::experimental::launch_policy policy{sycl_grid, sycl_block, launch_props, kernel_props};
+  auto event = compat::experimental::launch<cutlass::device_kernel<Kernel>, Kernel>(
+      policy, params.queue, static_cast<SparseAttnFwdParams>(params));
+}
+
+inline int sparse_mla_prefill_select_b_h(const XPUSparseAttnFwdParams& params) {
+  TORCH_CHECK(params.h_q > 1, "h_q must be > 1");
+  if (params.h_q < 8) return 8;
+  if (params.h_q < 16) return 16;
+  if (params.h_q < 32) return 32;
```
mingfeima:
```suggestion
  if (params.h_q <= 8) return 8;
  if (params.h_q <= 16) return 16;
  if (params.h_q <= 32) return 32;
```

decode is using <=. for 8/16/32 it will select 16/32/64 and introduce extra padding.

src/sycl/kernels/mla_sparse/sparse_mla_prefill_fwd.hpp
```diff
@@ -0,0 +1,813 @@
...
+  auto event = compat::experimental::launch<cutlass::device_kernel<Kernel>, Kernel>(
+      policy, params.queue, static_cast<SparseAttnFwdParams>(params));
+}
+
+template <int D_QK, bool HAVE_TOPK_LENGTH, bool HAS_ATTN_SINK, int B_H>
+void launch_mla_prefill_fwd_kernel_policy(const XPUSparseAttnFwdParams& params) {
+  using Traits = KernelTraits<B_H>;
+
+  TORCH_CHECK(params.h_kv == 1, "h_kv must be 1");
+  TORCH_CHECK(params.h_q > 1, "h_q must be > 1");
```
mingfeima: `sparse_mla_decode_fp8_fwd.hpp` L749-749

is doing
```
  TORCH_CHECK(params.h_q > 0, "h_q must be > 0");
```

also `sparse_mla_decode_fp8_fwd.hpp` L807


better to align.

and usually it is good practice to put these arguments checks in the entry function which is bind to python.

and skip all the tedious checks inside kernel files (unless necessay ones).

src/sycl/mla_sparse_decode.cpp
```diff
@@ -1,113 +1,273 @@
...
+  int b = q.size(0);
+  int s_q = q.size(1);
+  int h_q = q.size(2);
+  int d_qk = q.size(3);
+  int num_blocks = kv.size(0);
+  int page_block_size = kv.size(1);
+  int h_kv = kv.size(2);
+  int head_bytes = kv.size(3);
+  int topk = indices.size(2);
+  bool have_topk_length = topk_length.has_value();
```
mingfeima: HAVE_TOPK_LENGTH should be true if either topk_length or extra_topk_length

update test case as well, this leak is not covered in the UT.

python/sgl_kernel/attention.py
```diff
@@ -333,24 +336,147 @@ def flash_mla_with_kvcache(
                 and extra_topk_length.dtype == torch.int32
                 and extra_topk_length.shape == (B,)
             ), "extra_topk_length must be int32 on the same device with shape [B]"
-        torch.ops.sgl_kernel.flash_mla_sparse_decode.default(
-            out,
-            lse,
+        is_fp8_query = q.dtype == torch.float8_e4m3fn
```
mingfeima: we support fp8 in this patch? comment it if not. add a todo. seems a dead code.

tests/test_flash_mla_sparse_prefill.py
```diff
@@ -0,0 +1,158 @@
...
+"""
+
+import pytest
+import torch
+from sgl_kernel import flash_mla_sparse_prefill
+
+default_params = {
+    "has_attn_sink": [False, True],
+    "has_topk_length": [False, True],
+    "d_qk": [512],
```
mingfeima:
```suggestion
    "d_qk": [512, 576],
```

cmake has 576. also covers it.

src/sycl/kernels/mla_sparse/sparse_mla_decode_fp8_fwd.hpp
```diff
@@ -0,0 +1,836 @@
...
+        break;
+      case 32:
+        launch_sparse_mla_decode_fp8_fwd_kernel_policy<D_QK, HAS_TOPK_LENGTH, IS_FP8_QUERY, HAS_ATTN_SINK, 32>(params);
+        break;
+      default:
+        launch_sparse_mla_decode_fp8_fwd_kernel_policy<D_QK, HAS_TOPK_LENGTH, IS_FP8_QUERY, HAS_ATTN_SINK, 64>(params);
+        break;
+    }
+  });
+}
```
mingfeima: the recent heated discussion is that whether our code templated are too much..

here `HAS_TOPK_LENGTH` is a not necessary, you can check ptr == nullptr and it is one time thing.

you can let AI check whether remove it will affect the benchmark perf.

mingfeima (approved):
some minor changes are needed. generally LGTM.

python/sgl_kernel/attention.py
mingfeima: L293 checkss q.dtype == torch.bfloat16

jayachandranb-ai (changes requested):
While the perf improvement/feature additions from the new PR is very much welcome, the following are the concerns:

1. Structural Alignment:
The new PR removes  sycl-tla collective-builder layered structure(main-loop, epilogue etc) used in our repo for attention kernels.
(This is the same collective/kernel/device structure used by all our attention kernels viz. flash_attention_v2, non-sparse mla, 
original sparse mla etc.).
Instead the PR bundles the main code as mostly running code in two flat files.
So, 
i. There is no shared collective/epilogue/scheduler and no reuse of the MLA runner's split-KV / workspace mechanism.
ii. This also means if we have to support a new platform, the kernel body needs forking/if-else constructs and re-tuning instead of adding only necessary changes like DispatchPolicy specialization.
iii. This might also make the support of multiple kernel flavours if needed for any future unified kernel library difficult/patchy.
 
2. Runtime alignment :
The new PR differs in the way how the kernel is launched, how parameters are structured/passed, the runner path etc from what is currently aligned across our sycl-tla path and the CUDA reference.

 
In short, the design of the original sparse MLA had built in provisions for abstractions/modularity as per the current industry practices(CUTLASS/sycl-tla/CUDA style)
So the current PR should ideally have been built on top of the that design rather than completely removing those aspects.
Especially, let us keep the sycl-tla structure unless there is any compelling reason to avoid it and we want to avoid it in our existing kernels also.

src/sycl/kernels/mla_sparse/sparse_mla_prefill_fwd.hpp
Valentine233: I tried the suggested change, and the perf drops for certain shapes. The reason is that `B_H` will affect other configs, for example the subgroup layout. I've done a round of tuning for `B_H`, and the current strategy becomes:
```
if (params.h_q <= 8) return 8;
if (params.h_q <= 32) return 32;
```
With the change, we see no more perf drop, and the bandwidth (GB/s) improves from 55.95 to 68.03 for (seq_q=512, H=16, topk=2048).

For future work, a wider range of config tuning should be done.

src/sycl/kernels/mla_sparse/sparse_mla_prefill_fwd.hpp
Valentine233: Thanks, aligned both to `h_q>0`.

src/sycl/mla_sparse_decode.cpp
Valentine233: Thanks. `HAVE_TOPK_LENGTH` removed and UT added.

python/sgl_kernel/attention.py
Valentine233: Thanks and updated.

tests/test_flash_mla_sparse_prefill.py
Valentine233: Thanks and updated.

src/sycl/kernels/mla_sparse/sparse_mla_decode_fp8_fwd.hpp
Valentine233: Thanks. Removed `HAS_TOPK_LENGTH` for both decode and prefill, with benchmark validation.

mingfeima: haha, boss complained that our engineers over-templated ..

mingfeima: you can try harder, is this the only argument that can be removed? and how about the wheel package size increase? you can contact jianan how to do the wheel package size measurement.

src/sycl/kernels/mla_sparse/sparse_mla_prefill_fwd.hpp
mingfeima: this is the right thing to do.

src/sycl/kernels/mla_sparse/sparse_mla_decode_fp8_fwd.hpp
Valentine233: Also disabled `IS_FP8_QUERY` as DSV4 only uses BF16 query.
The wheel package size increases `1M`, which is acceptable as we add an additional sparse MLA prefill kernel.

Valentine233:
Landed https://github.com/sgl-project/sgl-kernel-xpu/pull/337.

PR #322, opened 2026-07-22
Updates for first release

ZailiWang:
- Update 3rd party notice file. The only hard dependency of `sglang-kernel-xpu` is `torch`.
- Update `.toml` file with the correct package name, version, et al.
- Update `version.py` with correct version number.


pyproject.toml
```diff
@@ -6,22 +6,21 @@ requires = [
...
 readme = "README.md"
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 license = { file = "LICENSE" }
 classifiers = [
   "Programming Language :: Python :: 3",
   "License :: OSI Approved :: Apache Software License",
-  "Environment :: GPU"
 ]
 dependencies = []
```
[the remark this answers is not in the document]
ZailiWang: `torch` dependency cannot be directly added here for now since in conda env the index-url of the dependent packages cannot be specified. Later when `uv` package manager is applied this can be achieved.

For the current release we have specified the `torch` installation explicitly in the doc & Dockerfile.

THIRDPARTYNOTICES.txt
```diff
@@ -1,488 +1,86 @@
-Notice for flashinfer-ai/flashinfer
+Notice for pytorch/pytorch
 -------------------------------
-                                 Apache License
-                           Version 2.0, January 2004
-                        http://www.apache.org/licenses/
```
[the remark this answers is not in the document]
ZailiWang: Other packages are indirect dependecies through pytorch.

pyproject.toml
```diff
@@ -6,22 +6,21 @@ requires = [
 build-backend = "scikit_build_core.build"
 
 [project]
-name = "sgl-kernel"
-version = "0.11.0"
-description = "Kernel Library for SGLang"
+name = "sglang-kernel-xpu"
```
polisettyvarma, line 9: better to remove xpu

pyproject.toml
```diff
@@ -6,22 +6,21 @@ requires = [
...
+version = "0.1.0"
```
polisettyvarma, line 10: adding version to xpu like torch is better

pyproject.toml
```diff
@@ -34,6 +33,6 @@ cmake.build-type = "Release"
 build-dir = "build"
 minimum-version = "build-system.requires"
 
-wheel.py-api = "cp39"
+wheel.py-api = "cp310"
```
polisettyvarma, line 39: any plan on manylinux support ?

pyproject.toml
ZailiWang: this is the agreed package name

pyproject.toml
ZailiWang: Yes, we have tested the validity of manaylinux support.
Also, the wheel can support py310-312, which was vefified.
The `py310` is in line with current CUDA sglang-kernel package: https://pypi.org/project/sglang-kernel/#files

mingfeima (approved):

mingfeima:
haha at last...

PR #328, opened 2026-07-23
add sycl kernel: plan_compress_prefill

jiayisunx:
### Motivation
Adds an SYCL implementation of the plan_compress_prefill planner and wires it into the Torch extension so Python can call the SYCL kernel instead of the previous pure-PyTorch implementation.

### Modifications
Register plan_compress_prefill in TORCH_LIBRARY_FRAGMENT(sgl_kernel, ...) for XPU dispatch.
Add SYCL kernel + C++ wrapper implementation for at::native::xpu::plan_compress_prefill.
Update the Python wrapper to call torch.ops.sgl_kernel.plan_compress_prefill instead of computing the plan via PyTorch ops.

### Performance
The SYCL implementation significantly improves the performance of plan_compress_prefill on XPU.
case | ratio | bs | seq_lens | extend_lens | num_q_tokens | sycl implementation time (us) | pure-PyTorch implementation time (us)
-- | -- | -- | -- | -- | -- | -- | --
c4_prefill_seq8_ext8_paged | 4 | 1 | 8 | 8 | 8 | 9.765 | 1185.589
c4_prefill_seq32_ext32_paged | 4 | 1 | 32 | 32 | 32 | 10.223 | 1173.261
c4_prefill_seq256_ext256_paged | 4 | 1 | 256 | 256 | 256 | 10.498 | 1221.63
c4_prefill_seq1024_ext1024_paged | 4 | 1 | 1024 | 1024 | 1024 | 17.304 | 1416.887
c128_prefill_seq128_ext128_paged | 128 | 1 | 128 | 128 | 128 | 7.471 | 857.95
c128_prefill_seq256_ext256_paged | 128 | 1 | 256 | 256 | 256 | 7.729 | 860.995
c128_prefill_seq512_ext512_paged | 128 | 1 | 512 | 512 | 512 | 9.198 | 887.474

### DIFF
--- a/src/sycl/CompressPlan.cpp
+++ b/src/sycl/CompressPlan.cpp
@@ -2,6 +2,8 @@
 #include <c10/xpu/XPUStream.h>
 #include <torch/all.h>
 
+#include <algorithm>
+#include <limits>
 #include <sycl/sycl.hpp>
 
 #include "SYCLHelpers.h"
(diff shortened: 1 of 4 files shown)


src/sycl/CompressPlan.cpp
```diff
@@ -0,0 +1,393 @@
...
+struct CompressPrefillStage0Kernel {
+  void operator()(sycl::nd_item<1> item) const {
+    if (item.get_global_id(0) != 0) {
+      return;
+    }
+
+    const bool is_overlap = (compress_ratio_ == 4);
+    const int32_t window_size = compress_ratio_ * (is_overlap ? 2 : 1);
+
+    uint32_t counter = 0;
```
polisettyvarma: cuda kernel supports MTP also, any plan to add that ?

src/sycl/CompressPlan.cpp
```diff
@@ -0,0 +1,393 @@
...
+    uint32_t counter_w = 0;
+
+    for (uint32_t batch_id = 0; batch_id < batch_size_; ++batch_id) {
+      const int32_t seq_len = static_cast<int32_t>(seq_lens_ptr_[batch_id]);
+      const int32_t extend_len = static_cast<int32_t>(extend_lens_ptr_[batch_id]);
+      const int32_t prefix_len = seq_len - extend_len;
+      const int32_t last_c_pos = (seq_len / compress_ratio_) * compress_ratio_;
+      const int32_t first_w_pos = sycl::min(last_c_pos - (is_overlap ? compress_ratio_ : 0), seq_len - mtp_pad_);
+
+      for (int32_t j = 0; j < extend_len; ++j) {
```
polisettyvarma: j is incremented by 1 but cuda kernel uses block_size
does that mean we are not vectorizing ?

src/sycl/CompressPlan.cpp
```diff
@@ -0,0 +1,393 @@
...
+        const int32_t position = prefix_len + j;
+        const uint32_t ragged_id = counter + static_cast<uint32_t>(j);
+
+        if (((position + 1) % compress_ratio_) == 0) {
+          const int32_t buffer_len = window_size - sycl::min(j + 1, window_size);
+          int32_t* plan = plan_c_i32_ + static_cast<int32_t>(counter_c) * 4;
+          plan[0] = position + 1;
+          plan[1] = ((buffer_len & 0xFFFF) << 16) | (static_cast<int32_t>(ragged_id) & 0xFFFF);
```
polisettyvarma: please cross check this in cuda kernel instead of pytorch code

src/sycl/CompressPlan.cpp
```diff
@@ -0,0 +1,393 @@
...
+          int32_t* plan = plan_w_i32_ + static_cast<int32_t>(counter_w) * 2;
+          plan[0] = ((static_cast<int32_t>(batch_id) & 0xFFFF) << 16) | (static_cast<int32_t>(ragged_id) & 0xFFFF);
+          plan[1] = position + 1;
+          ++counter_w;
+        }
+      }
+      counter += static_cast<uint32_t>(extend_len);
+    }
+
+    for (uint32_t k = counter_c; k < num_q_tokens_; ++k) {
```
polisettyvarma: same as above

src/sycl/CompressPlan.cpp
```diff
@@ -0,0 +1,393 @@
...
+      int32_t* plan = plan_c_i32_ + static_cast<int32_t>(k) * 4;
+      plan[0] = -1;
+      plan[1] = 0;
+      plan[2] = -1;
+      plan[3] = -1;
+    }
+    for (uint32_t k = counter_w; k < num_q_tokens_; ++k) {
```
polisettyvarma: same as above

src/sycl/CompressPlan.cpp
```diff
@@ -0,0 +1,393 @@
...
+          ++counter_w;
+        }
+      }
+      counter += static_cast<uint32_t>(extend_len);
+    }
+
+    for (uint32_t k = counter_c; k < num_q_tokens_; ++k) {
+      int32_t* plan = plan_c_i32_ + static_cast<int32_t>(k) * 4;
+      plan[0] = -1;
+      plan[1] = 0;
```
polisettyvarma: this is also -1 in cuda kernel, right ?

src/sycl/CompressPlan.cpp
```diff
@@ -0,0 +1,393 @@
...
+        const int32_t* mapping = r2t_ptr_ + rid * stride_r2t_;
+        int64_t raw_loc = static_cast<int64_t>(mapping[position]);
+        int64_t state_loc = f2s_ptr_[raw_loc];
+        write_loc = (state_loc / swa_page_size_) * ring_size_ + (state_loc % ring_size_);
+      }
+
+      plan_w[0] = packed & 0xFFFF;
+      plan_w[1] = static_cast<int32_t>(write_loc);
+    }
+  }
```
polisettyvarma, line 423: are we handling all cases like cuda kernel ?

src/sycl/CompressPlan.cpp
```diff
@@ -0,0 +1,393 @@
...
+  int32_t compress_ratio_;
+  int64_t stride_r2t_;
+  uint32_t num_work_;
+};
+
+}  // namespace CompressPlanImpl
+
+constexpr int64_t kPlanCBytes = sizeof(int32_t) * 4;
+constexpr int64_t kPlanWBytes = sizeof(int32_t) * 2;
+constexpr int64_t kPlanDBytes = sizeof(int32_t) * 4;
```
polisettyvarma: instead of magic numbers, can we follow like cuda struct style ?

polisettyvarma (changes requested):

src/sycl/CompressPlan.cpp
jiayisunx: Added.

src/sycl/CompressPlan.cpp
jiayisunx: Done, thanks for your comments!

src/sycl/CompressPlan.cpp
jiayisunx: Done.

src/sycl/CompressPlan.cpp
jiayisunx: There is one difference: CUDA has a dedicated CPU-input path, while on XPU I first move the inputs to XPU with .to(xpu), then run the same unified XPU kernel path.

mingfeima (commented):
Do we have test cases to cover this change?

jiayisunx:
> Do we have test cases to cover this change?

Yes, test cases were already added when the PyTorch implementation was introduced. (tests/deepseek_v4/test_c128_v2.py, tests/deepseek_v4/test_c4_v2.py)

polisettyvarma:
> > Do we have test cases to cover this change?
> 
> Yes, test cases were already added when the PyTorch implementation was introduced. (tests/deepseek_v4/test_c128_v2.py, tests/deepseek_v4/test_c4_v2.py)

during refactor deepseek_v4 mid folder is removed and we have UTs directly in tests now

mingfeima (approved):

src/sycl/CompressPlan.cpp
```diff
@@ -141,4 +496,138 @@ torch::Tensor plan_compress_decode(
...
+  const int32_t compress_ratio_i32 = static_cast<int32_t>(compress_ratio);
+  const int32_t swa_page_size_i32 = static_cast<int32_t>(swa_page_size);
+  const int32_t ring_size_i32 = static_cast<int32_t>(ring_size);
+
+  const uint32_t batch_size = static_cast<uint32_t>(seq_lens_xpu.numel());
+  constexpr uint32_t kMaxTokens = static_cast<uint32_t>(std::numeric_limits<uint16_t>::max());
+  constexpr uint32_t kMaxPrefillBatchSize = 1024;
+  TORCH_CHECK(compress_ratio_i32 == 4 || compress_ratio_i32 == 128, "compress_ratio must be 4 or 128");
+  TORCH_CHECK(
+      num_q_tokens_u32 >= batch_size && num_q_tokens_u32 <= kMaxTokens, "num_q_tokens must be in [batch_size, 65535]");
```
polisettyvarma, line 562: does cuda also have similar restriction ?

jiayisunx: Yes

PR #338, opened 2026-07-24
DeepSeek V4 Add support for fused_k_norm_rope_flashmla using torch impl

polisettyvarma:
UT fails as described in test file


mkumargarg (approved):

pralay-das (commented):
as implementation is okay, and in CUDA it is running fine, when sycl kernel will be available we will enable the test case and remove the torch function call.

PR #341, opened 2026-07-26
[MoE] Add Xe2 FP8 W8A16 grouped GEMM support

CaoE:
## Summary

Add FP8 E4M3 MoE support for Xe2/BMG with BF16 activations and outputs.

The existing `use_fp8_w8a8=True` API remains compatible, but dispatches to W8A16 because Xe2 has no native FP8 x FP8 MMA and the evaluated W8A8 path was slower.

## Changes

- Add Xe2 FP8 W8A16 grouped GEMM with:
  - per-expert scalar and 128x128 block scales
  - optional expert bias
  - SiLU, GELU, ReLU2, GPT-OSS SwiGLU, and DeepSeek-V4 gated activation
  - AOT and runtime JIT support

- Use a static grid-stride scheduler for grouped GEMMs with small routed
  work or short-K scalar scales:
  - `total_m <= num_experts`, or
  - scalar-scale GEMMs with `K <= 128`.
 
  This avoids dynamic tile-stealing atomics and improves short-GEMM latency
  by about 2.2%-3.6%.
- Add a specialized small-batch input-preparation path for FP8 MoE.
  For workloads with a small number of routed rows, one workgroup performs the expert histogram, prefix computation, stable route ranking, and activation scatter, avoiding the generic multi-kernel preparation path.

## Performance

Measured on Intel Arc Pro B60 with PyTorch 2.12.0+xpu using XPU events.

### Small-batch input preparation

| Shape `(M, TopK, E, H)` | Generic / Optimized time ratio |
|---|---:|
| `(1, 8, 256, 2048)` | 2.28x |
| `(2, 8, 128, 2048)` | 2.28x |
| `(4, 8, 128, 2048)` | 1.61x |
| `(5, 8, 128, 2048)` | 1.40x |
| `(8, 8, 128, 128)` | 2.26x |

#### Scalar FP8 weight scales

| Shape `(H, I, E, TopK)` | Tokens | BF16/FP8 time ratio |
|---|---:|---:|
| `(7168, 512, 256, 8)` | 1 | 1.76x |
| `(7168, 512, 256, 8)` | 32 | 2.15x |
| `(7168, 512, 256, 8)` | 2048 | 1.58x |
| `(2048, 256, 512, 10)` | 1 | 1.18x |
| `(2048, 256, 512, 10)` | 32 | 1.92x |
| `(2048, 256, 512, 10)` | 2048 | 1.38x |

#### 128x128 block FP8 weight scales

| Shape `(H, I, E, TopK)` | Tokens | BF16/FP8 time ratio |
|---|---:|---:|
| `(7168, 512, 256, 8)` | 1 | 1.73x |
| `(7168, 512, 256, 8)` | 32 | 2.15x |
| `(7168, 512, 256, 8)` | 2048 | 1.58x |
| `(2048, 256, 512, 10)` | 1 | 1.18x |
| `(2048, 256, 512, 10)` | 32 | 1.90x |
| `(2048, 256, 512, 10)` | 2048 | 1.32x |


python/sgl_kernel/moe.py
```diff
@@ -571,6 +655,49 @@ def fused_experts(
...
+            "use_fp8_w8a8=True currently requires hidden and intermediate "
+            "dimensions to be divisible by the FP8 group size 128"
+        )
+        assert (
+            w1.is_contiguous() and w2.is_contiguous()
+        ), "use_fp8_w8a8=True requires contiguous expert weights"
+        # Only 2-D block-quant (e.g. DeepSeek-style 128x128) scale tensors are
+        # supported in this first version; per-tensor (single scalar) weight
+        # scale is intentionally not wired up yet (see
+        # xpu_fp8_moe_minimal_plan.md - "do not try to unify every
```
polisettyvarma: this doesn't exist

src/sycl/kernels/moe/xe20/w8a16/moe_kernel.hpp
airMeng, line 1: can you reuse W4A16 files since they are the same?

src/sycl/kernels/moe/xe20/fp8/moe_mainloop.hpp
```diff
@@ -0,0 +1,412 @@
...
+      TiledMMA mma,
+      int thr_id,
+      BiasTensor Bias,
+      int gemm_n,
+      int m_actual) {
+    if constexpr (WeightScalePerExpert) {
+      run_w8a16_scalar(A, Bp, w_scale_gmem, w_scale_row_stride, D, blk_coord, mma, thr_id, Bias, gemm_n, m_actual);
+    } else {
+      constexpr int BLK_M = get<0>(decltype(mma.tile_mnk()){});
+      if (m_actual % BLK_M == 0) {
```
airMeng: Shall we request the FWK side to do the padding? You can compare the performance w/ and w/o padding

@jmunetong for awareness

src/sycl/kernels/moe/xe20/fp8/moe_mainloop.hpp
```diff
@@ -0,0 +1,412 @@
...
+              }
+            }
+          }
+        }
+      }
+      reorder(tArA_packed, tSrA);
+      reorder(tBrB_packed, tSrB);
+      CUTLASS_PRAGMA_UNROLL
+      for (int i = 0; i < tSrB.size(); ++i) {
+        tSrB(i) = apply_bf16_weight_scale(tSrB(i), w_scale);
```
airMeng: Can you try to reuse W4A16 code with several constexpr dispatch?

src/sycl/kernels/moe/xe20/w8a16/moe_kernel.hpp
CaoE: Thank you for your comments. I will try to reuse the w4a16 implementation.

src/sycl/kernels/moe/xe20/fp8/moe_mainloop.hpp
CaoE: Thanks for your comments. Use real m size instead. No significant performance changes were observed w/ and w/o padding (~2% grouped gemm time).

src/sycl/kernels/moe/xe20/w8a16/moe_kernel.hpp
CaoE: I evaluated sharing the W4A16 kernel/mainloop with W8A16. Although their outer GEMM structure is similar, their weight processing differs significantly:

W8A16 scalar scales are applied to FP32 accumulators after the K-loop.
W8A16 block scales uniformly transform each FP8 B fragment.
W4A16 requires INT4/MXFP4 unpacking, per-channel/per-group scales, optional zero-point handling, and scale prefetching before MMA.

Technically, reusing W4A16 files is feasible, but a forced reuse involves significant changes and requires adding extra abstractors, which doesn't save much code. Furthermore, it doesn't reduce the required AOT/JIT specializations.

Both W4A16 and W8A16 are still evolving. Sharing their performance-critical code now would couple future tuning and make changes to one format more likely to affect the other, increasing maintenance and regression risk. Therefore, I prefer separate implementations while reusing only stable, format-independent utilities.

python/sgl_kernel/moe.py
```diff
@@ -635,42 +758,57 @@ def fused_experts(
...
         out_hidden_states = torch.empty_like(hidden_states)
 
-    topk_ids = topk_ids.int() if topk_ids.dtype == torch.long else topk_ids
     expert_offsets = _get_moe_ws(
         "expert_offsets", (E,), torch.int32, hidden_states.device
     )
-    problem_sizes1 = _get_moe_ws(
-        "problem_sizes1", (E, 3), torch.int32, hidden_states.device
+    use_small_prepare = (
+        _should_use_small_moe_prepare(M, TopK, hidden_dims, E) and use_fp8_weight
```
airMeng, line 767: Whether it will help for other dtypes? I am interested since moe permute is a common bottleneck

src/sycl/kernels/moe/xe20/w8a16/moe_kernel.hpp
```diff
@@ -0,0 +1,192 @@
...
+#include "../common/xe_gemm.hpp"
+#include "cute/tensor.hpp"
+#include "cutlass/cutlass.h"
+#include "cutlass/float8.h"
+#include "cutlass/gemm/gemm.h"
+#include "cutlass/gemm/group_array_problem_shape.hpp"
+#include "cutlass/gemm/kernel/tile_scheduler.hpp"
+#include "cutlass/kernel_hardware_info.hpp"
+#include "cutlass/platform/platform.h"
+#include "cutlass/util/packed_stride.hpp"
```
airMeng, line 21: can you help to remove useless headers? We didn't check whether these are necessary.
Also apply for MoE of other data types

airMeng (approved):

python/sgl_kernel/moe.py
CaoE: This should help with other types as well; I will test it further.

src/sycl/kernels/moe/xe20/w8a16/moe_kernel.hpp
CaoE: Removed unnecessary header files on both bf16 and fp8 moe.

PR #347, opened 2026-07-28
Add SYCL support for flash_compress128_decode

jiayisunx:
### Motivation
Adds an SYCL implementation of `flash_compress128_decode `and wires it through the C++/Torch extension so the Python API uses the native op instead of a Torch fallback implementation.

### Modifications
Register `flash_compress128_decode` as an XPU op in the SYCL torch extension.
Add a new SYCL kernel implementation for `flash_compress128_decode`.
Centralize compress plan struct definitions into a shared SYCL header (`Compress.h`) and update callers.

### Performance
The SYCL implementation significantly improves the performance of `flash_compress128_decode `on XPU.
case | bs | seq_lens | head_dim | sycl implementation time(us) | pure-PyTorch implementation time(us) | speedup (PyTorch/SYCL)
-- | -- | -- | -- | -- | -- | --
c128_seq128_hd256_f32_paged | 1 | 128 | 256 | 26.042 | 598.282 | 22.97x
c128_seq256_hd256_f32_paged | 1 | 256 | 256 | 26.025 | 503.406 | 19.34x
c128_seq512_hd256_f32_paged | 1 | 512 | 256 | 25.955 | 531.319 | 20.47x
c128_seq128_hd512_f32_paged | 1 | 128 | 512 | 26.284 | 509.533 | 19.39x
c128_seq256_hd512_f32_paged | 1 | 256 | 512 | 26.313 | 544.807 | 20.71x
c128_seq512_hd512_f32_paged | 1 | 512 | 512 | 26.343 | 512.693 | 19.46x
c128_seq128_hd1024_f32_paged | 1 | 128 | 1024 | 26.260 | 515.029 | 19.61x
c128_seq256_hd1024_f32_paged | 1 | 256 | 1024 | 26.250 | 547.651 | 20.86x
c128_seq512_hd1024_f32_paged | 1 | 512 | 1024 | 26.176 | 512.585 | 19.58x

### DIFF
--- a/src/sycl/FlashCompress128.cpp
+++ b/src/sycl/FlashCompress128.cpp
@@ -42,12 +42,12 @@ struct FlashCompress128DecodeKernel {
       kv_dst[head_dim_ + h] = static_cast<buffer_t>(kv_src[head_dim_ + h]);
     }
 
-    item.barrier(sycl::access::fence_space::global_and_local);
-
     // Compress only when a 128-token chunk is completed.
     if ((plan.write_loc % 128) != 127) {
       return;
     }
+
+    item.barrier(sycl::access::fence_space::global_and_local);
 
     // Last split may be partial when head_dim is not divisible by kTileDim.
     if (h >= head_dim_) {
--- a/src/sycl/FlashCompress128.cpp
+++ b/src/sycl/FlashCompress128.cpp
@@ -1,0 +1,138 @@
+#include <ATen/ATen.h>
+#include <c10/xpu/XPUStream.h>
+#include <torch/all.h>
+
+#include <limits>
+#include <sycl/sycl.hpp>
+
+#include "Compress.h"
+#include "Utils.h"
+
+namespace at::native::xpu {
+
+namespace {
+
+template <typename buffer_t, typename input_t, typename out_t>
+struct FlashCompress128DecodeKernel {
+  void operator()(sycl::nd_item<1> item) const {
+    const uint32_t bid = static_cast<uint32_t>(item.get_global_id(0));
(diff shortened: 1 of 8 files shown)


src/sycl/FlashCompress128.cpp
```diff
@@ -0,0 +1,162 @@
...
+      kv_dst[h] = static_cast<buffer_t>(kv_src[h]);
+      kv_dst[head_dim_ + h] = static_cast<buffer_t>(kv_src[head_dim_ + h]);
+    }
+
+    item.barrier(sycl::access::fence_space::global_and_local);
+
+    // Compress only when a 128-token chunk is completed.
+    if ((plan.write_loc % 128) != 127) {
+      return;
+    }
```
mingfeima: The work-group barrier runs on every decode step, but only the compress path (write_loc % 128 == 127) needs to observe the just-written row. Please move the compress check above the barrier (all lanes share the same plan) so non-compress steps skip the sync.

mingfeima (changes requested):
`tests/test_c128_v2.py`: existing decode coverage via test_prefill_then_decode is good and CI-green; optional follow-up to parametrize head_dim beyond 512 (and/or bf16) to match the shapes in the PR bench.

src/sycl/FlashCompress128.cpp
jiayisunx: Done, thanks for the comments!

jiayisunx: Hi @mingfeima , could you please review it again? Thanks!

mingfeima (approved):

PR #352, opened 2026-07-29
Add optional workspace to MOE and GDN arg list

Xia-Weiwen:
Using caller-provided workspace can reduce the number of calls to `torch.empty`, thus reducing the size of reserved memory on device, which reduces the risk of OOM.
If workspaces are not provided, it falls back to `torch.empty`, which does not break any existing use cases.
Test cases are also added for MOE and GDN with workspaces.

Benchmark data

Latency | Before | After
-- | -- | --
Prefill E2E | 3.12087 sec | 3.11497 sec
Decode E2E | 0.04200 sec | 0.04200 sec
GDN prefill | 211.235 ms | 210.980 ms
GDN decode | 1.055 ms | 1.056 ms

Note: The intention of this PR is solve the OOM issue, not for performance gain though it should bring tiny performance benefit.


python/sgl_kernel/moe.py
```diff
@@ -332,6 +332,43 @@ def cutlass_fp4_group_mm(
...
+    for d in shape:
+        numel *= d
+    cur = workspace.get(name)
+    if cur is None or cur.numel() < numel or cur.dtype != dtype or cur.device != device:
+        # Growing replaces the previous buffer. Without a sync here, an in-flight kernel
+        # from an earlier call may still be reading/writing that old buffer's memory,
+        # causing a data race. With the headroom, growth only happens a handful of times
+        # total, so this sync is effectively free amortized over a server's lifetime.
+        if cur is not None:
+            torch.xpu.synchronize(device)
```
airMeng: The comments doesn't convince me...

@claude check here

airMeng: @copilot+gpt-5.6-sol:max check here also

src/sycl/gdn_attn/gdn_attn_interface_impl.hpp
```diff
@@ -207,6 +211,26 @@ void gdn_attention(
   auto& queue = dpcppGetCurrentQueue();
   auto dtype = projected_states_qkvz.dtype();
   auto device = projected_states_qkvz.device();
+
+  // Reuse a caller-provided persistent workspace buffer to reduce allocations.
+  // Falls back to the original allocate-per-call behavior when no workspace is
+  // supplied or it is too small / wrong dtype, so existing callers keep working.
+  auto make_ws_tensor =
```
mingfeima: If this is going to be called every run?

if so, we need to remove the overheads from `narrow` and `zero_`

* you can use the `shape` to do the narrow inside sycl kernel without explicitly chunk the ws into vector of tensors
* the zero_ can be done in sycl kernel as well, no need to call t.zero_()

mingfeima (changes requested):

src/sycl/gdn_attn/gdn_attn_interface_impl.hpp
mingfeima: similar to how flash_attn handles the `cu_seqlen`.

python/sgl_kernel/moe.py
Xia-Weiwen: Thanks. It's fixed by copilot.

src/sycl/gdn_attn/gdn_attn_interface_impl.hpp
Xia-Weiwen: Thanks for reviewing. After experiments, I removed `narrow` but still leave the `zero_` here, because
- The original code calls `aten::zeros` so the overhead is not introduced by this PR.
- I tried to fuse `zero` into the kernel but performance became lower. Analysis showed that the reason is `tensorzero_` has better kernel launch grid configuration than the GDN kernel. So, separating them results in better performance.

After all, the `make_ws_tensor` replaces `torch.empty` and `torch.zeros`, which should reduce the total overhead.

mingfeima:
@Xia-Weiwen do you have bench numbers? And profile results breakdown (you can pick 1 or 2 shape in the benchmark and do op level breakdown)?

src/sycl/gdn_attn/gdn_attn_interface_impl.hpp
mingfeima: Yes i understand. 5 points added! do you want 10? ~_~

it is OK that you left is this way, mark a TODO (future possible optimizations) at the top of the file, catching up with other's logic is often more diffult.

Xia-Weiwen: The TODO is already there. Thanks.

Xia-Weiwen:
> @Xia-Weiwen do you have bench numbers? And profile results breakdown (you can pick 1 or 2 shape in the benchmark and do op level breakdown)?

Benchmark data are added in the PR description. Thanks.

python/sgl_kernel/gdn_attn.py
```diff
@@ -1,7 +1,79 @@
...
+    num_decodes: int,
+    non_spec_token: int,
+    batch_size: int,
+    num_k_heads: int,
+    num_v_heads: int,
+    head_k_dim: int,
+    head_v_dim: int,
+    tp_size: int,
+    dtype: torch.dtype,
+) -> int:
```
mingfeima, line 27: from what i see this logic is like:

* `_gdn_workspace_bytes_needed`  - this one computes nbytes needed (from python)
* `make_ws_tensor` - chop tensors (from C++)

and it has two padding logic, which are duplicated:

```
### this is python
_GDN_WS_ALIGN = 256
_align_up(nbytes)  # (nbytes + 255) // 256 * 256

### this is C++
constexpr int64_t kWsAlign = 256;
aligned_offset = (ws_cursor_bytes + kWsAlign - 1) / kWsAlign * kWsAlign;
```

i prefer to put `_gdn_workspace_bytes_needed` in C++ as well and expose interface to python, therefore you can reduce the duplicated logic and much easier to maintain.

mingfeima (changes requested):
another thing is

change `std::optional<torch::Tensor>& workspace` to `std::optional<torch::Tensor>& workspace`

include/sgl_kernel_ops.h
```diff
@@ -944,7 +944,9 @@ void gdn_attention(
     const std::optional<torch::Tensor>& num_accepted_tokens,
     const int64_t num_actual_tokens,
     const int64_t tp_size,
-    const bool reorder_input);
+    const bool reorder_input,
+    // Optional pre-allocated scratch buffer: a single flat torch::kUInt8 tensor.
+    const std::optional<torch::Tensor>& workspace);
```
mingfeima, line 949: one tensor is much better ! thumbs up!

src/sycl/gdn_attn/gdn_attn_interface_impl.hpp
```diff
@@ -207,6 +208,43 @@ void gdn_attention(
...
+  int64_t ws_capacity_bytes = ws_active ? workspace->numel() : 0;
+  int64_t ws_cursor_bytes = 0;
+  auto make_ws_tensor = [&](const std::vector<int64_t>& shape, torch::ScalarType st, bool zero_init) -> torch::Tensor {
+    int64_t numel = 1;
+    for (auto d : shape)
+      numel *= d;
+    auto opts = torch::dtype(st).device(device).requires_grad(false);
+    if (ws_active) {
+      int64_t nbytes = numel * static_cast<int64_t>(c10::elementSize(st));
+      int64_t aligned_offset = (ws_cursor_bytes + kWsAlign - 1) / kWsAlign * kWsAlign;
```
mingfeima:
```suggestion
      iint64_t aligned_offset = div_up(ws_cursor_bytes, kWsAlign) * kWsAlign;
```

Xia-Weiwen: Updated. Thanks.

python/sgl_kernel/gdn_attn.py
Xia-Weiwen: Thanks. Now a new utility op is added to compute the workspace size so that the logic can be reused in both C++ and Python.

mingfeima (approved):

PR #357, opened 2026-07-29
[sycl] Support plain row-major UE8M0 scale output in per-token-group …

msinnha1:
The per_token_group_quant_8bit kernels (v1 and v2) previously only emitted UE8M0 (float8_e8m0fnu) block scales in the column-major, uint32-packed / swizzled layout used by the fused GEMM path. The plain row-major branch hard-asserted `!SCALE_UE8M0`, so callers that need a contiguous `[tokens, K/group]` uint8 scale (e.g. torch._scaled_mm with the BlockWise1x32 recipe, which expects an unswizzled per-block scale) could not use the fused activation-quant kernel and had to fall back to a pure-torch quantizer.

This adds a plain row-major UE8M0 code path in both kernels:

  * Kernel row-major branch: write the UE8M0 byte contiguously at the group index via scale_packed_t=uint8_t (scale_element_t=uint8_t), instead of static_asserting it away. Semantically identical to the existing float row-major store, just byte-typed.
  * Dispatch: when scale_ue8m0 is requested on a row-major (non column-major) output, instantiate the kernel with <..., IS_COLUMN_MAJOR=false, SCALE_UE8M0=true, scale_packed_t=uint8_t> and cast output_s to uint8_t*. The v2 launch passes the packed type explicitly so it does not default to uint32_t.

Existing paths are unchanged: column-major uint32-packed UE8M0 and plain float scales dispatch exactly as before. No public API / signature change -- callers select this path simply by passing a uint8 `[tokens, K/group]` scale tensor with scale_ue8m0=true.

Tests: add test_plain_row_major_ue8m0 (v1) and test_plain_row_major_ue8m0_v2 covering this layout, which was previously unreachable (the existing scale_ue8m0 helper asserts column-major + tma-aligned). Each checks the emitted UE8M0 byte per block is exactly ceil(log2(amax/448)) + 127 and that the dequant round-trip cosine >= 0.99.

Verified: clang-format + pre-commit clean; icpx -fsycl compiles both kernels for the intel_gpu_cri device; the v1 test passes on the CRI simulator across 5 shapes; and the reference byte math matches a bit-faithful CPU emulation of the kernel.

### DIFF
--- a/src/sycl/per_token_group_quant_8bit.cpp
+++ b/src/sycl/per_token_group_quant_8bit.cpp
@@ -111,8 +111,11 @@ struct PerTokenGroupQuant8bitKernel : public __SYCL_KER_CONFIG_CONVENTION__ {
       scale_output = reinterpret_cast<scale_element_t*>(output_s) +
                      (col_idx * scale_stride * num_elems_per_pack + row_idx * num_elems_per_pack + pack_idx);
     } else {
-      static_assert(!SCALE_UE8M0);
-      scale_output = output_s + global_group_id;
+      // Plain row-major scale layout. Supports float scales (non-UE8M0) and, when
+      // the kernel is instantiated with scale_packed_t=uint8_t, the UE8M0 byte
+      // written contiguously at the group index -- a plain [tokens, K/group] uint8
+      // scale (consumed directly by e.g. torch._scaled_mm BlockWise1x32).
+      scale_output = reinterpret_cast<scale_element_t*>(output_s) + global_group_id;
     }
 
     using vec_type = vec_t<T, VEC_SIZE>;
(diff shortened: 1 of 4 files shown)


kareemshaik80 (commented):
Trigger CI. Overall LGTM.

sspintel:
It might be better to cherry-pick the changes for v1 quant kernel from [#207](https://github.com/intel-innersource/frameworks.ai.pytorch.sgl-kernel-xpu/pull/207/changes#diff-cedfa832df6a20b0d2e95ec3e7ab8f010094402c8530328832edadbcb8946c51) on innersource-main so that we have code consistency between innersource and upstream, and to avoid any future conflicts.

mkumargarg (approved):

PR #371, opened 2026-08-05
[XPU][MiniMax] Add decode block-top-k kernels

Amrutha-M05:
## Summary

Port the MiniMax-M3 decode block-top-k kernels from CUDA to Intel XPU as AOT-compiled SYCL. Used by the MiniMax sparse decode indexer to pick the top-k block ids from per-(head, batch) score rows.

- **`minimax_decode_topk`** — block-id output `[num_heads, batch, topk]`, front-packed and `-1` padded. Drop-in for the Triton 2-stage top-k fallback.
- **`minimax_decode_topk_page_table`** — fused top-k + page-table transform for the dense backend (`trtllm_mha` / `fa3`); emits per-(batch, kv_head) page rows, effective KV length, and head-encoded page indices for DP attention.

Mirrors the CUDA source at `sglang/python/sglang/kernels/jit/csrc/minimax/minimax_decode_topk.cuh`, same output contract apart from PDL (Hopper-only), which is dropped. Both ops are registered on `torch::kXPU` via `TORCH_LIBRARY_FRAGMENT` and built into `common_ops`, so no compiler is needed on the runtime host and there is no first-call compile latency. `block_size`, `topk` and `page_size` stay runtime arguments, so one prebuilt extension covers every shape.

## Changes

| File | |
|---|---|
| `src/sycl/MinimaxDecodeTopK.cpp` | Both kernels and both `SGL_KERNEL_EXPORT` entry points (836 lines). Templated on the `seq_lens` width, instantiated for `int32`/`int64`. |
| `include/sgl_kernel_ops.h` | Declare the two ops. |
| `src/torch_extension_sycl.cc` | `m.def` schemas + `m.impl(..., torch::kXPU, ...)`. |
| `python/sgl_kernel/minimax.py`, `__init__.py` | `torch.ops` wrappers; API matches `sglang.kernels.ops.attention.minimax_decode_topk`. |
| `tests/test_minimax_decode_topk.py`, `tests/run_suite.py` | 16 accuracy tests vs an inline PyTorch reference, registered for CI. |
| `benchmark/bench_minimax_decode_topk.py`, `benchmark/run_suite.py` | Bandwidth-vs-peak sweeps, registered in the `per-commit` suite (15.5 s). |

Both kernel functors derive from `__SYCL_KER_CONFIG_CONVENTION__`, declare SLM in `sycl_ker_config_convention()`, and launch through `sycl_kernel_submit()`.

`topk_forward` has a trivial fast path (`num_blocks <= topk`) plus three regimes:

| regime | bound | strategy |
|---|---|---|
| small | `num_blocks <= 128` | rank-by-compare, no radix |
| register-1 | `<= 512` | 4-pass 8-bit radix, 1 element/thread in a register |
| register-M | `<= 4096` | 4-pass 8-bit radix, 8 elements/thread, `uint32_t` liveness bitmask |

Sub-group pinned to 32 lanes via `[[sycl::reqd_sub_group_size(32)]]` so the `kNumWarps=16` × 32-lane histogram fan-out and register-M slicing match the CUDA CTA layout.

## Performance

Arc Pro B60 (peak **456 GB/s**), oneAPI 2026.1, torch 2.14.0.dev+xpu, `block_size=64 page_size=16 topk=16`.

These are selection kernels — comparisons and radix bucketing, no multiply-accumulate — so there is no meaningful FLOP count; bandwidth is the only useful roofline. Bytes counted are the traffic actually touched per shape, with the score row excluded in the trivial regime where it is never read.

**`minimax_decode_topk`** — block-id output, `num_heads=32`

| regime | num_blocks | batch | time (us) | GB/s | % peak |
|:---|---:|---:|---:|---:|---:|
| trivial | 8 | 1 / 16 / 64 | 5.57 / 8.91 / 20.36 | 0.37 / 3.69 / 6.45 | 0.1 / 0.8 / 1.4 |
| small | 128 | 1 / 16 / 64 | 7.66 / 28.02 / 90.05 | 2.41 / 10.53 / 13.10 | 0.5 / 2.3 / 2.9 |
| register-1 | 512 | 1 / 16 / 64 | 9.87 / 50.21 / 172.40 | 6.85 / 21.54 / 25.09 | 1.5 / 4.7 / 5.5 |
| register-M | 4096 | 1 / 16 / 64 | 26.77 / 204.43 / 754.19 | 19.66 / 41.20 / **44.66** | 4.3 / 9.0 / **9.8** |

**`minimax_decode_topk_page_table`** — fused top-k + page table, `num_heads=8`

| regime | num_blocks | batch | time (us) | GB/s | % peak |
|:---|---:|---:|---:|---:|---:|
| trivial | 8 | 1 / 16 / 64 | 6.51 / 9.14 / 18.41 | 0.32 / 3.65 / 7.24 | 0.1 / 0.8 / 1.6 |
| small | 128 | 1 / 16 / 64 | 10.78 / 19.11 / 49.27 | 0.76 / 6.89 / 10.69 | 0.2 / 1.5 / 2.3 |
| register-1 | 512 | 1 / 16 / 64 | 11.90 / 24.90 / 70.47 | 1.72 / 13.19 / 18.63 | 0.4 / 2.9 / 4.1 |
| register-M | 4096 | 1 / 16 / 64 | 23.59 / 77.29 / 252.08 | 5.73 / 27.99 / **34.33** | 1.3 / 6.1 / **7.5** |

**`topk` sensitivity** — `num_blocks=1024`, `batch=16`, `num_heads=32`

| topk | 1 | 4 | 16 | 32 |
|---|---:|---:|---:|---:|
| time (us) | 85.18 | 85.68 | 88.59 | 89.95 |
| GB/s | 24.64 | 24.57 | 24.04 | 24.04 |

Cost is flat in `topk` (1 -> 32 is +5.6% time): the radix passes are driven by `num_blocks`, not `k`, so `topk` only changes the final write width. That lets the indexer raise `topk` without a latency cliff.

**Reading the % peak.** It rises monotonically with `batch` and `num_blocks`, topping out near 45 GB/s. The low absolute fraction is expected, not a regression:

1. Not a streaming kernel — each row is loaded once, then goes through up to four radix passes of cross-lane scan/reduce plus SLM atomics. Runtime is dominated by that reduction work, so effective bandwidth understates utilisation.
2. One work-group per (head, batch) row — at `batch=1` that is 32 work-groups (block-id) or 8 (page-table), which cannot fill the device. The ramp with batch is occupancy, not bandwidth.

So the large-batch `register-M` rows are the meaningful roofline; `trivial`/`small` are launch-latency-bound. The page-table kernel trails because it runs the same selection over 4x fewer heads (`kv_heads=8` vs `qo_heads=32`), then adds a `req_to_token` gather per emitted page.

## CUDA → SYCL mapping

| CUDA | SYCL |
|---|---|
| warp (32 lanes) | sub-group pinned via `[[sycl::reqd_sub_group_size(32)]]` |
| `__shfl_up_sync` (inc-scan) | `inclusive_scan_over_group(sg, val, plus)` |
| warp reduce-sum | `reduce_over_group(sg, val, plus)` |
| `atomicAdd` (shared) | `atomic_ref<..., local_space, relaxed>` |
| `__syncthreads()` | `item.barrier(local_space)` |
| `__shared__` struct | `local_accessor<TopKSmem, 1>` / `<PageTableSmem, 1>` in `sycl_ker_config_convention()` |
| kernel launch | `sycl_kernel_submit()` (`src/sycl/SYCLHelpers.h`) |
| `__float_as_uint` | `sycl::bit_cast<uint32_t>` |
| PDL / PDLWaitPrimary | dropped (no XPU equivalent) |

## Test plan

- [x] `pytest tests/test_minimax_decode_topk.py -v` on Arc Pro B60 — **16 passed**.
- [x] All four paths: trivial, small, register-1, register-M. Both output modes. Both `seq_lens` dtypes, so both instantiations.
- [x] Multi-kv-head DP layout, `page_size > 1`, permuted `req_to_token`, wrapped slot ids, caller-provided `out`.
- [x] Both ops resolve from a plain `import sgl_kernel`, no compiler on the host.
- [x] Neither launcher calls `.wait()` — kernels stay async on the current XPU stream.
- [x] Moving to `sycl_kernel_submit()` is codegen-neutral: the rebuilt `.so` is byte-identical to the pre-refactor build.

CC: @siju-samuel @rbabukv


kareemshaik80 (commented):
please add benchmark tests

mingfeima:
@airMeng please help review this one!

dependency for https://github.com/sgl-project/sglang/pull/37193

include/sgl_kernel/minimax/minimax_decode_topk.hpp
airMeng, line 1: why put the kernel implementations in the headers, move to src/sycl

benchmark/bench_minimax_decode_topk.py
```diff
@@ -0,0 +1,422 @@
...
+        "score": score,
+        "seq_lens": seq_lens,
+        "out": out,
+        "block_size": block_size,
+        "topk": topk,
+        "max_seqblock": max_seqblock,
+    }
+
+
+def _torch_topk_block_ids(state):
```
airMeng: remove the eager implementation, we should compare the performance with HW spec directly

benchmark/bench_minimax_decode_topk.py
```diff
@@ -0,0 +1,422 @@
...
+    print("=" * 80)
+    print(df.to_markdown(index=False))
+
+    _report_speedup(
+        df[df["kernel"] == "minimax_decode_topk"],
+        ["num_blocks", "batch", "case"],
+        "case",
+        "minimax_decode_topk",
+    )
+    _report_speedup(
```
airMeng: we do not need to compare with eager, report effective BW/TFlops directly

benchmark/bench_minimax_decode_topk.py
airMeng, line 1: add the file into run_suite.py

airMeng (changes requested):

airMeng:
removed from https://github.com/sgl-project/sglang/pull/37193

kareemshaik80 (commented):
LGTM!

benchmark/bench_minimax_decode_topk.py
airMeng: https://github.com/sgl-project/sgl-kernel-xpu/blob/main/benchmark/run_suite.py I mean here

src/sycl/MinimaxDecodeTopK.cpp
```diff
@@ -0,0 +1,836 @@
...
+            static_cast<const SeqLenT*>(seq_lens),
+            static_cast<int32_t*>(topk_idx),
+            batch,
+            num_heads,
+            max_seqblock,
+            block_size,
+            topk,
+            smem));
+  });
+  // Do NOT .wait(): this runs on the current XPU stream and PyTorch owns synchronization.
```
airMeng: remove the comments

src/sycl/MinimaxDecodeTopK.cpp
```diff
@@ -0,0 +1,836 @@
...
+  int32_t num_heads_;
+  int32_t max_seqblock_;
+  int32_t block_size_;
+  int32_t topk_;
+  int32_t page_size_;
+  int32_t r2t_stride_;
+  int32_t max_kv_len_;
+  int32_t max_reqs_;
+  int32_t max_sparse_pages_;
+  ::sycl::local_accessor<PageTableSmem, 1> smem_;
```
airMeng, line 593: Hi, for kernels needing smem, please use the unified launcher https://github.com/sgl-project/sgl-kernel-xpu/blob/3f3c73216dc978eeda689c5960c61a549fc309c2/src/sycl/SYCLHelpers.h#L46-L52

You can find an example usage like https://github.com/sgl-project/sgl-kernel-xpu/blob/3f3c73216dc978eeda689c5960c61a549fc309c2/src/sycl/HCHead.cpp#L64

src/sycl/MinimaxDecodeTopK.cpp
```diff
@@ -0,0 +1,836 @@
...
+            block_size,
+            topk,
+            page_size,
+            r2t_stride,
+            max_kv_len,
+            max_reqs,
+            max_sparse_pages,
+            smem));
+  });
+  // Do NOT .wait(): this runs on the current XPU stream and PyTorch owns synchronization.
```
airMeng: remove comments

airMeng (commented):
Overall LGTM except some minor comments

airMeng (approved):

airMeng:
@Amrutha-M05 Can you share the performance measurements results(BW/Tflops) in the PR description?

sakotkar (approved):

PR #378, opened 2026-08-07
Add SYCL support for compress_norm_rope_store

jiayisunx:
### Motivation
Adds native SYCL/XPU implementations and bindings for the fused norm+RoPE+store path and extends coverage for additional cache layouts, moving Python wrappers toward direct torch.ops dispatch in this codebase.

### Modifications
Adds a new SYCL kernel implementation for `fused_norm_rope_store `(`head_dim `128/512) with optional indexer preshuffle and FlashMLA bf16-store modes.
Exposes `compress_norm_rope_store ` via torch_extension_sycl.cc and updates Python wrappers to call it.
Expands test_fused_norm_rope_v2.py to validate preshuffle and bf16-store cache layouts.

### Performance
The SYCL implementation significantly improves the performance of fused_norm_rope_store on XPU.
ase | head_dim | num_tokens | dtype | sycl implementation time(us) | pure-PyTorch implementation time(us) | speedup (PyTorch/SYCL)
-- | -- | -- | -- | -- | -- | --
fnrs_hd128_n1_bf16_fp8 | 128 | 1 | bf16 | 8.030 | 2652.963 | 330.49x
fnrs_hd128_n16_bf16_fp8 | 128 | 16 | bf16 | 8.577 | 2676.044 | 312.03x
fnrs_hd128_n96_bf16_fp8 | 128 | 96 | bf16 | 8.982 | 2987.199 | 332.68x
fnrs_hd512_n1_bf16_fp8 | 512 | 1 | bf16 | 9.877 | 2446.216 | 247.75x
fnrs_hd512_n16_bf16_fp8 | 512 | 16 | bf16 | 9.817 | 2337.794 | 238.16x
fnrs_hd512_n96_bf16_fp8 | 512 | 96 | bf16 | 13.955 | 2693.499 | 193.00x


tests/test_fused_norm_rope_v2.py
```diff
@@ -230,6 +240,33 @@ def _dequant_indexer(cache: torch.Tensor, num_tokens: int) -> torch.Tensor:
...
+                token_tile_id * (preshuffle_size * _IDX_VALUE_BYTES)
+                + col_tile_id * (preshuffle_size * preshuffle_size)
+                + token_in_tile * preshuffle_size
+                + col_in_tile
+            )
+            packed[d] = cache[page, value_offset]
+        fp8 = packed.view(torch.float8_e4m3fn)
+        scale = cache[page, sbase : sbase + _IDX_SCALE_BYTES].view(torch.float32)
+        vals.append(fp8.float() * scale)
+    return torch.stack(vals)
```
mingfeima, line 298: `_dequant_indexer_preshuffle` is definitely not an efficient implementation.

it is OK if this part is a minor part. otherwise sycl it.

src/sycl/FusedNormRopeStore.cpp
```diff
@@ -0,0 +1,689 @@
...
+    for (uint32_t mask = kSubGroupSize; mask <= 32; mask <<= 1) {
+      const uint32_t peer = lid ^ mask;
+      const float o0 = data[static_cast<int64_t>(peer)];
+      const float o1 = data[static_cast<int64_t>(peer) + kIndexerBlockSize];
+      h0 = (lid & mask) ? (o0 - h0) : (h0 + o0);
+      h1 = (lid & mask) ? (o1 - h1) : (h1 + o1);
+      data[d0] = h0;
+      data[d1] = h1;
+      item.barrier(sycl::access::fence_space::local_space);
+    }
```
mingfeima, line 252: cross subgroup (mask=16/32) loads peer SLM then stores in-place with no barrier between load and store.

data race?

tests/test_fused_norm_rope_v2.py
```diff
@@ -150,6 +158,8 @@ def _run(head_dim: int, num_tokens: int):
         compress_ratio=plan.compress_ratio,
         page_size=PAGE_SIZE,
         use_fp4=False,
```
mingfeima: do we need to check use_fp4 in [True, False]

src/sycl/FusedNormRopeStore.cpp
mingfeima: the following is what Kimi K3 tells me, you can reference :)

```
**[blocking]** `src/sycl/FusedNormRopeStore.cpp:242-250`: The cross-subgroup
WHT (`mask=16/32`) reads `data[peer]` / `data[peer + kIndexerBlockSize]` and
writes `data[d0]` / `data[d1]` in-place in the same loop body, with only a
trailing `item.barrier(local_space)`. `lid^16/32` points into a different
subgroup, which may already have issued its store before this work-item's
loads complete, so the butterfly can read post-store values. The single barrier
at L250 breaks the loop-carried dep between iterations, but not the
within-iteration load-vs-store ordering; CUDA avoids this with warp `shfl_xor`.
Fix: add a second barrier inside the loop before the loads (or save the current
values to SLM, barrier, then compute):
    const uint32_t peer = lid ^ mask;
    const float o0 = data[...];            // peer loads
    const float o1 = data[...];
    const float c0 = (lid & mask) ? (o0 - h0) : (h0 + o0);
    const float c1 = (lid & mask) ? (o1 - h1) : (h1 + o1);
    item.barrier(sycl::access::fence_space::local_space);  // all loads done
    data[d0] = c0;
    data[d1] = c1;
    item.barrier(sycl::access::fence_space::local_space);
```

mingfeima (changes requested):

tests/test_fused_norm_rope_v2.py
jiayisunx: This helper is a test-only reference implementation for correctness checks, not a production path. I think we prioritize readability here.

mingfeima: sooooooory i did not notice that it is test function :(

src/sycl/FusedNormRopeStore.cpp
jiayisunx: Done, thanks!

tests/test_fused_norm_rope_v2.py
jiayisunx: Added, thanks for your comments!

mingfeima (approved):

PR #388, opened 2026-08-12
Remove unnecessary reshape of output for Neox style ROPE

Xia-Weiwen:
The reshape here (adding a leading `1` dimension) is wrong here because
1. The kernel uses the pointer, unrelated to the shape.
2. The shape has nothing to do the the algorithm.
3. It silently violates the "RoPE preserves tensor shape" invariant.

It's most plausibly a copy/paste leftover from some other kernel path in the codebase that legitimately deals with a  [1, batch, ...] -shaped tensor layout (e.g., a batched-request wrapper elsewhere), mistakenly attached to the `is_neox_style` conditional instead of being removed — dead code that just happened to never get exercised by the test suite (which only tests `is_neox_style=False` for the 3D path) until Llama's standard rope path hit it.

A test case is added to cover this fix.


airMeng (approved):

mingfeima (approved):
good catch!

PR #389, opened 2026-08-12
Optimize Gemma4 FMHA decode Split-K path

yuankuns:
## Summary
- Optimize the Gemma4 paged decode Split-K path for `B=1`, `Hq=8`, `Hkv=1`, and `Dv=512`.
- Use a 256-wide output tile for HD512 and parallelize Split-K reduction.
- Fix Split-K variable-length shape and paged-KV cache extent handling.
- Keep HD512 on Split-K below the 64-page threshold and use the Gemma4-specific split-count heuristic.

## Validation
- Default-feature clean wheel build passed (2013 targets).
- GPU 1 BF16 reference correctness passed for 12 contexts from 1 through 5120 tokens.

## GPU 1 Measurement
BF16 paged KV, `B=1/Hq=8/Hkv=1/D=512`, automatic Split-K, XPU event median, 18 runs x 2000 iterations.

| Context | Baseline `03637f08` | This PR |
| ---: | ---: | ---: |
| 4096 | 768.6 us / 10.9 GB/s | 31.1 us / 269.8 GB/s |
| 4097 | 1079.3 us / 7.8 GB/s | 31.1 us / 270.1 GB/s |
| 4608 | 1192.9 us / 7.9 GB/s | 31.6 us / 298.4 GB/s |
| 5120 | 1323.3 us / 7.9 GB/s | 31.8 us / 329.6 GB/s |

## Gemma4 TP=4 E2E Benchmark
SGLang main `69a3c54c`, GPU `4,5,6,7`, TP=4, 4096 input tokens + 1024 output tokens, concurrency 1. The primary decode metric is median ITL.

| Metric | Baseline `03637f08` | This PR `acf148dc` | Change |
| --- | ---: | ---: | ---: |
| Median ITL | 96.44 ms | 87.00 ms | 1.108x / -9.79% |
| Mean TPOT | 92.97 ms | 87.42 ms | 1.064x / -5.98% |
| P90 ITL | 105.99 ms | 92.08 ms | -13.12% |
| TTFT | 1475.52 ms | 1470.68 ms | -0.33% |
| E2E latency | 96.59 s | 90.90 s | 1.063x / -5.89% |
| Output throughput | 10.60 tok/s | 11.26 tok/s | +6.27% |

The operator-level gain is diluted in whole-model decode by GEMM, MoE, TP communication, and other non-FMHA work.

### DIFF
--- a/tests/test_flash_attention.py
+++ b/tests/test_flash_attention.py
@@ -978,7 +978,7 @@ def test_flash_attn_kvcache(
     [torch.bfloat16, torch.float16]
     + ([torch.float8_e4m3fn] if not DISABLE_FP8 else []),
 )
-@pytest.mark.parametrize("nheads_q,nheads_kv", [(16, 16), (16, 4), (16, 1)])
+@pytest.mark.parametrize("nheads_q,nheads_kv", [(16, 16), (16, 4), (16, 1), (8, 1)])
 @pytest.mark.parametrize("new_kv", [False])
 @pytest.mark.parametrize("causal", [False])
 @pytest.mark.parametrize("local", [True, False])
@@ -1006,9 +1006,18 @@ def test_flash_attn_kvcache(
 @pytest.mark.parametrize(
     "seqlen_k",
     [
-        128,
+        1,
+        63,
+        64,
+        65,
+        129,
+        512,
         1024,
+        4033,
         4096,
+        4097,
+        4608,
+        5120,
         8192,
     ],
 )
@@ -1035,8 +1044,10 @@ def test_flash_attn_decode_kvcache(
 ):
     from sgl_kernel.flash_attn import flash_attn_with_kvcache
 
-    if page_size is not None and seqlen_k % page_size != 0:
-        pytest.skip("page_size must divide seqlen_k")
+    if (nheads_q, nheads_kv) == (8, 1) and (
+        batch_size != 1 or d != 512 or page_size != 64
+    ):
+        pytest.skip("Gemma4 Split-K coverage only targets B=1, D=512, page_size=64")
     if seqlen_q > seqlen_k and new_kv:
         pytest.skip("new_kv requires seqlen_q <= seqlen_k")
     if not new_kv and rotary_fraction > 0.0:
@@ -1349,7 +1360,15 @@ def test_flash_attn_decode_kvcache(
         sin = sin.to(dtype) if sin is not None else None
         k_cache_saved = k_cache.clone() if page_size is None else k_cache_paged.clone()
         v_cache_saved = v_cache.clone() if page_size is None else v_cache_paged.clone()
-        num_splits_vals = [1, 0] if not DISABLE_SPLIT else [1]
+        num_splits_vals = (
+            [1, 0]
+            if batch_size == 1
+            and nheads_q == 8
+            and nheads_kv == 1
+            and d == 512
+            and page_size == 64
+            else [1]
+        )
         precompute_metadata_vals = [False]
         for num_splits, precompute_metadata in itertools.product(
             num_splits_vals, precompute_metadata_vals
--- a/python/sgl_kernel/flash_attn.py
+++ b/python/sgl_kernel/flash_attn.py
@@ -272,7 +272,7 @@ def flash_attn_with_kvcache(
         cu_seqlens_q,
         cu_seqlens_k,
         max_seqlen_q,
-        1,
+        max_seqlen_k,
         page_table,
         cache_batch_idx,
         cache_leftpad,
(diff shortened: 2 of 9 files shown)


benchmark/bench_fmha_decode.py
airMeng, line 1: can't reuse the current benchmarks?

cmake/BuildFlags.cmake
```diff
@@ -48,6 +48,10 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     list(APPEND SYCL_HOST_FLAGS -g -fno-omit-frame-pointer -O0)
   elseif(CMAKE_BUILD_TYPE MATCHES RelWithDebInfo)
     list(APPEND SYCL_HOST_FLAGS -g -O2)
+  elseif(CMAKE_BUILD_TYPE MATCHES MinSizeRel)
+    list(APPEND SYCL_HOST_FLAGS -Os -DNDEBUG)
+  else()
+    list(APPEND SYCL_HOST_FLAGS -O3 -DNDEBUG)
```
airMeng: I think we don't enable CMAKE_BUILD_TYPE=MinSizeRel in pip?

src/sycl/comm/common.h
```diff
@@ -60,4 +60,32 @@ static void launch(typename Kernel::Params params) {
...
+      syclex::sub_group_size<cute::intel::sg_size>, intelex::grf_size<GrfSize>};
+
+  launch_policy policy{
+      compat::dim3(grid.x, grid.y, grid.z),
+      compat::dim3(block.x, block.y, block.z),
+      launch_props,
+      kernel_props};
+  auto stream = at::xpu::getCurrentXPUStream();
+  auto q = stream.queue();
+  compat::experimental::launch<cutlass::device_kernel<Kernel>, Kernel, false>(policy, q, params);
```
airMeng: Any performance gains without event? If yes, can we make it as the default path.

airMeng (approved):

airMeng:
I wanted to click ```comment``` not approve and seems can't withdraw :)

tests/test_flash_attention.py
```diff
@@ -1006,9 +1006,18 @@ def test_flash_attn_kvcache(
...
+        64,
+        65,
+        129,
+        512,
         1024,
+        4033,
         4096,
+        4097,
+        4608,
+        5120,
```
airMeng, line 1020: Is it necessary to add so many tests? Might slow down the CI

tests/test_flash_attention.py
```diff
@@ -1349,7 +1360,15 @@ def test_flash_attn_decode_kvcache(
...
         v_cache_saved = v_cache.clone() if page_size is None else v_cache_paged.clone()
-        num_splits_vals = [1, 0] if not DISABLE_SPLIT else [1]
+        num_splits_vals = (
+            [1, 0]
+            if batch_size == 1
+            and nheads_q == 8
+            and nheads_kv == 1
+            and d == 512
+            and page_size == 64
+            else [1]
```
airMeng, line 1370: any reason why we only test split with this config?

tests/test_flash_attention.py
yuankuns: It's always a balance between more test / bug leak actually. I usually will consider those edge cases in my side so add them in the test if required.

tests/test_flash_attention.py
yuankuns: They are only enabled case, and will benefit the performance otherwise the tradeoff are bigger, and this is the same reason why split k was off by default.

jmunetong:
I have been able to recreate the results for this PR on gemma-4-31b. 

Both sides were measured on this host: the reproduced baseline ran against a from-source
non-editable install of `03637f08` (§6.1), the reproduced PR side against the PR head.

| Metric | PR: Baseline `03637f08` | PR: This PR `acf148dc` | **Baseline `03637f08` reproduced** | **PR389 reproduced** | Same-host A/B |
| --- | ---: | ---: | ---: | ---: | ---: |
| Median ITL | 96.44 ms | 87.00 ms | **105.14 ms** | **77.18 ms** | **1.36×** / −26.6% |
| Mean TPOT | 92.97 ms | 87.42 ms | **106.44 ms** | **82.68 ms** | **1.29×** / −22.3% |
| P90 ITL | 105.99 ms | 92.08 ms | **111.51 ms** | **104.20 ms** | 1.07× / −6.6% |
| TTFT | 1475.52 ms | 1470.68 ms | **1506.67 ms** | **1468.06 ms** | 1.03× / −2.6% |
| E2E latency | 96.59 s | 90.90 s | **110.39 s** | **86.05 s** | **1.28×** / −22.0% |
| Output throughput | 10.60 tok/s | 11.26 tok/s | **9.27 tok/s** | **11.90 tok/s** | **1.28×** / +28.4% |

With both sides measured here, decode improves on every metric: median ITL 105.14 → 77.18 ms (**1.36×**), mean
TPOT 106.44 → 82.68 ms (1.29×), E2E 110.39 → 86.05 s (1.28×), output throughput 9.27 → 11.90
tok/s (+28.4%). TTFT barely moves (1506.67 → 1468.06 ms, −2.6%), exactly as it should for a
change confined to the decode Split-K path.

jmunetong (approved):

PR #391, opened 2026-08-13
JIT compile wheel size 21MB

sunjiweiswift:
# RFC: Device-Aware Runtime JIT for Multi-Arch XPU Kernels

- **Status:** Draft (core + MoE landed; FMHA/MLA/GDN pending)
- **Area:** `sgl-kernel-xpu` runtime JIT (`src/jit/`) · Intel XPU (SYCL/`icpx`)
- **Date:** 2026-08-20

## Summary

The JIT already ships kernels as templates (`USE_SYCL_JIT`) instead of AOT
binaries, shrinking the wheel **~180 MB → ~22 MB**. But it compiles for one
architecture (Xe20/BMG) via a process-global `-fsycl-targets`. This RFC makes the
JIT **device-aware**: each launch classifies its `sycl::queue`'s device and
compiles/caches the correct per-arch variant, so **Xe20 (BMG)** and **Xe35
(XE3P)** can coexist.

## Motivation

- **Correctness:** hard-coded `intel_gpu_bmg_g21` produces wrong code/ISA on Xe35.
- **Multi-device:** one process may hold Xe20 *and* Xe35 queues; each launch must
  pick the matching variant.
- **Size:** AOT multi-arch = a fat binary (~2× instances); JIT builds only the
  arch(es) present — the strongest argument *for* JIT.

## Design

**1. Classify device (SYCL side, `Utils.h`)** — map the real architecture enum to
a stable code, mirroring `query_device`:
```cpp
static inline int jit_arch_code(at::DeviceIndex dev = -1) {
  switch (get_device_architecture(dev)) {
    case syclex::architecture::intel_gpu_bmg_g21:
    case syclex::architecture::intel_gpu_bmg_g31:
      return (int)sgl::jit::Arch::BMG;
    // case <xe35>: return (int)sgl::jit::Arch::XE3P;
    default: return (int)sgl::jit::Arch::BMG;
  }
}
```

**2. Arch profile (`jit/jit_arch.{h,cpp}`, torch/SYCL-free)**
```cpp
enum class Arch { BMG = 0, XE3P = 1 };
struct ArchProfile { std::string target, macro, suffix; };
// BMG  -> {"",              "",                 "bmg"}
// XE3P -> {"<xe35 id TBD>", "XPU_ENABLED_XE3P", "xe3p"}
```

**3. Per-spec target (engine)** — `CompileSpec.target` added;
`default_sycl_flags()` no longer hard-codes the target; `get_or_compile` appends
`-fsycl-targets=<spec.target | default>`, so it lands in the cache key.

**4. Thread `arch` through each op** — `resolve()`/`pack_key()` take `arch`, set
`spec.target`, add `-D<macro>`, append `_<suffix>` to the name + front-cache key.
Dispatcher passes `jit_arch_code()`.

**5. Multi-device dispatch** — arch is derived per launch and is part of the key:

| Launch | arch | key | artifact |
|---|---|---|---|
| GPU0 Xe20 | BMG | `(config, BMG)` | `..._bmg.so` |
| GPU1 Xe35 | XE3P | `(config, XE3P)` | `..._xe3p.so` (`-DXPU_ENABLED_XE3P`) |

Same config compiles twice; two `.so` coexist. Without arch in the key an Xe20
fn-ptr could be handed to an Xe35 queue — the bug this fixes.

## Why not one SPIR-V

Xe35 source diverges at compile time (`#if defined(XPU_ENABLED_XE3P)`: 512-GRF,
256 subgroups, 384 KiB SLM, different tiles). SPIR-V only finalizes ISA for the
*same* source; it can't recover a different code path. → compile per arch.

## Rollout & Compatibility

Landed: engine target + `jit_arch.*` + `Utils.h` + MoE (`GroupGemmXe20/W4A16`).
Pending: FMHA/MLA/GDN (identical threading), then Xe35 bring-up (fill
`ArchProfile{XE3P}.target`, the enum case, per-arch tiles). **BMG-only today is a
no-op** (arch resolves to 0; same flags/keys) — verified by a clean
`USE_SYCL_JIT=ON` build.

## Open Questions

1. Xe35 `-fsycl-targets` id and its `syclex::architecture` enum.
2. Do `-spirv-ext` flags vary per arch?
3. Per-arch tile heuristics in `resolve()` or `ArchProfile`?


src/jit/sycl_template_jit.cpp
```diff
@@ -0,0 +1,385 @@
...
+      }
+    }
+    out += tmpl[i];
+    ++i;
+  }
+  return out;
+}
+
+void* get_or_compile(const CompileSpec& spec, const JitConfig& config, std::string* err) {
+  auto set_err = [&](const std::string& m) {
```
airMeng, line 207: check the host memory for whether enough for compilation

mingfeima:
@sunjiweiswift could you please give everyone  a headsup about this, maybe prepare a simple slice on next weekly meeting:

* new files added
* old kernel files
* how to coop with CRI
* how to enable jit compile / how to enable aot compile
* how vllm is working on this
* resullts

and so on ...

kareemshaik80 (commented):
UT coverage is missing for several JIT-enabled kernels. It would be helpful to add tests validating JIT correctness against the corresponding AOT implementations, along with representative JIT vs AOT performance benchmark results.

kareemshaik80 (commented):
How sglang runtime invokes the JIT kernels? The interface is not clear. Do you import from sgl_kernel.jit.*? 
Do you call a high-level function or a low-level C-dispatch layer?
What parameters does the user pass, and how does the JIT decide which config to compile?
What happens if a config is unsupported or compilation fails?  Is it aligned with this PR https://github.com/sgl-project/sglang/pull/28716 for runtime interface.

sunjiweiswift:
> How sglang runtime invokes the JIT kernels? The interface is not clear. Do you import from sgl_kernel.jit.*? Do you call a high-level function or a low-level C-dispatch layer? What parameters does the user pass, and how does the JIT decide which config to compile? What happens if a config is unsupported or compilation fails?  Is it aligned with this PR [sgl-project/sglang#28716](https://github.com/sgl-project/sglang/pull/28716) for runtime interface.

Both AOT and JIT modes use the same flash_attn_with_kvcache interface.

The selection between JIT and AOT is controlled by the build flag USE_SYCL_JIT:

When USE_SYCL_JIT is enabled, the system operates in JIT mode. Kernel code is packaged into a shared library (.so file); at runtime, the system selects the appropriate template instantiation based on the actual configuration and compiles it.
When USE_SYCL_JIT is disabled, AOT mode is used, and the required template instantiations are generated during build time.

JIT and AOT share the same C++ implementation. As a result, after warmup, their runtime performance is identical. The only difference is that in JIT mode, each template instantiation incurs a one-time compilation overhead when it is used for the first time.

@mingfeima  @airMeng  pls review

airMeng (approved):

PR #342, opened 2026-07-27
[XPU] Optimize paged FMHA prefill for head dim 512

yuankuns:
## Summary

- port the paged FMHA score-reuse optimization for `HEAD_DIM=512` from CUTLASS-SYCL
- store scaled QK scores in a block2D workspace and reuse them for the second output tile
- preserve remainder-K masking for non-aligned KV lengths, including local-window attention
- keep the existing bf16-only paged prefill dispatch unchanged
- add `FMHA_PREFILL_ENABLE_SCORE_BLOCK2D_512`, enabled by default
- cap the score workspace at 1 GiB by slicing the batch and reusing the same workspace

## Correctness

Tested bf16 on Intel Arc Pro B60 with reverse page tables.

`B=1, SQ=127, SK=250, Hq/Hkv=4/2, D=512`:

| mask | max error | bad elements |
| --- | ---: | ---: |
| full | 0.001964 | 0 |
| causal | 0.002082 | 0 |
| local `(63,17)` | 0.003994 | 0 |

Also tested `B=2` with KV lengths `[250, 193]`, reverse page tables, and local `(63,17)` attention: max error `0.003724`, bad elements `0`.

Additional integration validation:

- full default-feature build: `1485/1485`
- `tests/test_flash_attention.py`: `1725 passed, 660 skipped`
- K-tail, causal GQA, and score-prefetch boundary cases: max absolute error `0.000174-0.000364`, no elements above `0.01`

### Original kernel case

Intel Arc Pro B60, bf16, page size 64, reverse page table, `B=1, SQ=512, SK=4096, Hq/Hkv=16/8, D=512`:

| workload | baseline | optimized | improvement |
| --- | ---: | ---: | ---: |
| full attention | 2.754 ms | 2.203 ms | 20.0% |
| Gemma4 global attention (causal) | 2.615 ms | 2.138 ms | 18.2% |

### Large production shape

Intel Arc Pro B60, bf16, page size 128, `B=16, SQ=SK=4096, Hq/Hkv=16/16, D=512`. The baseline is the saved pre-PR wheel; both wheels used the same benchmark script on GPU 1 with `triton.testing.do_bench`, `warmup=1000 ms`, and `rep=5000 ms`.

| workload | baseline | optimized | latency improvement | baseline TOPS | optimized TOPS |
| --- | ---: | ---: | ---: | ---: | ---: |
| full attention | 297.427 ms | 175.081 ms | 41.1% (`1.70x`) | 29.57 | 50.24 |
[...]

PR #319, opened 2026-07-22
[fmha] Split decode/prefill kernels into per-layout, per-dtype Template units

sunjiweiswift:
Split the large `decode.so` file to reduce memory usage during compilation.

PR #392, opened 2026-08-14
[XPU] Fix paged FMHA prefetch bounds and expand coverage

yuankuns:
## Summary

Fix the paged FMHA prefill final-tile page-table bounds issue and expand opt-in XPU FMHA coverage.

This page-table out-of-bounds bug was discovered recently through the expanded test coverage, but it is a pre-existing issue rather than a regression from PR #342. Earlier UT coverage did not repeatedly exercise the final-K-tile/page-boundary combination, so it remained undetected. The same failure reproduces on PR #319 and PR #252. The bounds fix is now included here.

- Clamp the speculative next K-tile index before page-table translation in the non-ScoreBlock2D paged prefill path. This prevents a final-iteration `K + 1` page-table access beyond the valid cache range.
- Extend the BF16/FP16 paged KV-cache matrix to head dimensions `64/96/128/192/256/512`.
- Focus the FP8 broad matrix on FP8 dtype and descale coverage, while retaining causal decode/prefill smoke coverage and long-sequence causal canaries.
- Add non-paged varlen K-tail coverage with non-power-of-two lengths.

The extended matrix remains opt-in via `FLASH_ATTENTION_KVCACHE_EXTENDED_TESTS=1`.

## Validation

- The six original paged/GQA hang repro shapes passed reference validation on GPUs 1, 2, and 3.
- Full extended suite on GPU 0: `32003 passed, 1896 skipped in 633.34s`.
- `black-jupyter --files tests/test_flash_attention.py` passes.

## Notes

The page-table bounds fix addresses the confirmed final-K-tile out-of-bounds path. The earlier resource-pressure experiment that held multiple K-block score fragments is not present in this branch.

### DIFF
--- a/src/sycl/kernels/flash_attention_v2/collective/xe_fmha_fwd_mainloop.hpp
+++ b/src/sycl/kernels/flash_attention_v2/collective/xe_fmha_fwd_mainloop.hpp
@@ -507,7 +507,10 @@ struct FMHAFwdMainloop<
 
       page_idx = next_page_idx;
       if constexpr (!ScoreBlock2D) {
-        next_page_idx = K + 1;
+        // The final iteration has no next K prefetch, but translating K + 1
+        // still dereferences the page table. Clamp it to the last valid tile
+        // so page_size=128 does not read one entry past the page table.
+        next_page_idx = cute::min(K + 1, k_end - 1);
         if constexpr (PagedKV) {
           next_page_idx = get_physical_k_tile(next_page_idx, l_coord, seq_len_kv_cache);
         }
--- a/tests/test_flash_attention.py
+++ b/tests/test_flash_attention.py
@@ -66,9 +66,7 @@ DISABLE_PACKGQA = True
 DISABLE_FP16 = True
 DISABLE_FP8 = True
 
-EXTENDED_KVCACHE_TESTS = (
-    os.getenv("FLASH_ATTENTION_KVCACHE_EXTENDED_TESTS") == "1"
-)
+EXTENDED_KVCACHE_TESTS = os.getenv("FLASH_ATTENTION_KVCACHE_EXTENDED_TESTS") == "1"
 KVCACHE_BATCH_SIZES = [5]
 KVCACHE_HEAD_CONFIGS = [(16, 16), (16, 4), (8, 1)]
 KVCACHE_SEQLEN_CONFIGS = [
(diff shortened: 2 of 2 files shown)


Valentine233:
@yuankuns Thanks! Recently, you mention that several PRs cause hang issues: https://github.com/sgl-project/sgl-kernel-xpu/pull/342, https://github.com/sgl-project/sgl-kernel-xpu/pull/384, https://github.com/sgl-project/sgl-kernel-xpu/pull/252, https://github.com/sgl-project/sgl-kernel-xpu/pull/319. This PR only fixes https://github.com/sgl-project/sgl-kernel-xpu/pull/342, and the rest three ones still exist. For the rest unresolved cases, could you please indicate in the description the corresponding guilty commit?

yuankuns:
> @yuankuns Thanks! Recently, you mention that several PRs cause hang issues: #342, #384, #252, #319. This PR only fixes #342, and the rest three ones still exist. For the rest unresolved cases, could you please indicate in the description the corresponding guilty commit?

The GPU hang is caused by a latent out-of-bounds page-table access in the paged-KV FMHA kernel. The issue was observed in the code from PR #252 and PR #319, both of which predate PR #342. PR #342 did not modify the page-table traversal logic and did not introduce this bug.

The issue was first exposed when new UT coverage expanded the GQA ratio to 8. For specific combinations such as GQA ratio = 8, page size = 128, and hd = 512, the kernel can prefetch a next page-table entry after reaching the final valid  K tile. This accesses an entry beyond the valid page-table range. The invalid value  may come from adjacent or unmapped memory, and if it is subsequently used as a  physical-page address, it can cause a GPU hang or device loss.

After expanding coverage further, we observed similar hangs for additional shapes,  including FP8 with head dimension 256, bf16 with hd=256 in code from both PR #252 and PR #319. This confirms that the problem is a pre-existing latent boundary bug rather than a regression introduced by PR #342. The previous UT matrix did not cover all relevant combinations of GQA ratio, page size, head dimension, KV-cache dtype, query/cache length, and final-page boundaries, so these cases were not triggered earlier.

The fix bounds the next page-table prefetch to the final valid K tile, preventing the out-of-range page-table access. With the fix applied, all previously observed hanging shapes pass on the current HEAD without GPU hangs or device loss.

airMeng (approved):

Valentine233 (approved):
Thanks for the fixing!
Several things to confirm:
1. Is `FLASH_ATTENTION_KVCACHE_EXTENDED_TESTS` added to reduce CI time?
2. Do this fix all the issues of other PRs I listed above?

yuankuns:
`FLASH_ATTENTION_KVCACHE_EXTENDED_TESTS` is opt-in to keep the default CI suite practical. It is disabled by default; it can be enabled in CI or run manually when broader regression coverage is needed.

For the listed PRs: the final-K-tile page-table OOB is a pre-existing FMHA issue that reproduces on #252 and #319, and was not introduced by #342. The fix in this PR addresses that shared page-table traversal bug; the previously observed six hang shapes now pass on current HEAD.

I also checked #384. It only changes build/profiling flags and does not modify FMHA or page-table traversal, so it is unrelated to this kernel issue.

PR #398, opened 2026-08-20
Improve fp8_paged_mqa_logits

Xia-Weiwen:
Added a chunking strategy to avoid OOM with large batch.
Improved performance by vectorization, tuning tile/workgroup size, etc.

### DIFF
--- a/src/sycl/Fp8MqaLogitsXe20.cpp
+++ b/src/sycl/Fp8MqaLogitsXe20.cpp
@@ -25,6 +25,8 @@ limitations under the License.
 #include <c10/xpu/XPUStream.h>
 #include <torch/all.h>
 
+#include <cstdlib>
+
 #include "kernels/nsa/fp8_mqa_gemm_xe20.hpp"
 #include "kernels/nsa/fp8_mqa_logits_kernel.hpp"
 #include "sgl_kernel_export.h"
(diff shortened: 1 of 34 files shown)


polisettyvarma (commented):
helps in OOM and which model, right ?

Xia-Weiwen:
> helps in OOM and which model, right ?

It can probably resolve the OOM issue of DeepSeek. Jianan is going to have a try.

src/sycl/Fp8MqaLogitsXe20.cpp
```diff
@@ -167,57 +169,99 @@ SGL_KERNEL_EXPORT torch::Tensor fp8_paged_mqa_logits(
...
+          " msl=",
+          msl,
+          " -> chunk_b=",
+          chunk_b,
+          " (",
+          cdiv(B_next, chunk_b),
+          " chunks)");
+    }
+
+    for (int start = 0; start < B_next; start += chunk_b) {
```
airMeng, line 207: avoid launching kernels in the loop, fuse into a larger kernel instead

Xia-Weiwen: Thanks for the suggestion. My agent says the loop introduces limited kernel launch overhead (~2% in the DeepSeek V4 case). I have added a TODO here to indicate an opportunity for optimizations. I think we can keep the current implementation to resolve the OOM issue which is blocking.

Xia-Weiwen: After an offline discussion, we decide to improve the kernel implementation for better performance so I have converted this PR to draft and will work on it later. Thanks.

Xia-Weiwen: The chunking strategy is applied on the batch. The gather - GEMM - reduction kernels are launched one by one. And since they have different launch configs, it's hard to fuse them.
The performance of these kernels is also improved.

src/sycl/kernels/nsa/fp8_mqa_logits_kernel.hpp
```diff
@@ -145,17 +145,25 @@ struct PagedKGatherKernel {
...
+  // 16 bytes (4x uint32) per transaction instead of 4 bytes.
+  using Vec4 = sycl::vec<uint32_t, 4>;
+
   void operator()(sycl::nd_item<2> item) const {
     int b = item.get_global_id(0);
     int kj = item.get_global_id(1);
     if (b >= B || kj >= max_seq_len) return;
 
     int out_idx = b * max_seq_len + kj;
+    int n_vec4 = D / 16;
```
airMeng: 16 is sg_size?

tests/test_fp8_mqa_logits.py
```diff
@@ -374,5 +379,69 @@ def test_fp8_paged_mqa_logits_3d_input():
...
+    Forces a tiny per-call chunk budget via the env var, in a fresh
+    subprocess: the C++ side caches the budget in a function-local
+    `static const` on first use, so setting the env var in-process would
+    have no effect once any earlier test in this file has already invoked
+    the op (as several do, with the default budget). With the chosen shape
+    (B=8, H=256, D=128, page_size=max_seq_len=128) and a 1MiB budget,
+    per_batch_bytes=144128 so chunk_b=7, i.e. two unequal chunks (7 + 1).
+    """
+    tests_dir = os.path.dirname(os.path.abspath(__file__))
+    script = f"""
```
airMeng: suggest to use some more smart LLM :)

airMeng (changes requested):

src/sycl/kernels/nsa/fp8_mqa_logits_kernel.hpp
Xia-Weiwen: Thanks. It's the vec len. Replaced with a constexpr var.

tests/test_fp8_mqa_logits.py
Xia-Weiwen: Thanks. This part is polished.

airMeng (approved):

PR #401, opened 2026-08-22
[XPU] Fix ScoreBlock2D workspace OOM

yuankuns:
## Summary
- allocate the ScoreBlock2D workspace as bytes instead of dtype-sized elements
- slice large score workspaces across batches, GQA query heads, and Q-tile chunks while preserving enough Q-tile parallelism for BMG/CRI
- preserve original strides when rebasing sliced varlen tensors
- add extended paged and non-paged OOM regressions across large sequence, high-batch/low-head, and nested batch/head slicing cases

## Performance and workspace-cap trade-off
- `FMHA_SCORE_WS_CAP_MB` is a memory/performance knob (default: 1024 MiB). A larger score buffer lets each launch cover more Q tiles, reducing the number of ScoreStore/ScoreLoad launch pairs and their global score-buffer write/read overhead. The cost is higher peak XPU memory and a greater OOM risk.
- At `B=1, Hq/Hkv=16/2, Sq=Sk=4096, D=512` (BF16 non-paged varlen prefill), the full Q extent fits in 512 MiB: `query_tile_slice=32`, one ScoreStore/ScoreLoad pair, 51.3 ms vs. 72.3 ms without ScoreBlock2D (29.1% lower latency; 41.0% higher throughput).
- At `Sq=Sk=32768`, the 1 GiB cap allows only `query_tile_slice=8`: 32 Q-tile chunks / ScoreStore+ScoreLoad pairs. ScoreBlock2D still improves latency, 3992.4 ms vs. 5004.5 ms (20.2% lower latency; 25.4% higher throughput), but the extra launch and buffer-traffic cost reduces the realized gain. Raising the cap would reduce the chunk count, subject to available memory.
- The optimization is not a universal win for tiny/low-parallelism prefill shapes: `Sq=1` showed no material benefit, and tested `Hq<=4, Sq=16/64` cases regressed. These measurements are specific to the HD512 BF16 non-paged varlen path.

## Validation
- `FLASH_ATTENTION_KVCACHE_EXTENDED_TESTS=1 pytest -q tests/test_flash_attention.py`: 35256 passed, 4996 skipped
- large ScoreBlock2D OOM regressions: 6 passed
- Valentine 32K case: `B=1, Hq/Hkv=16/2, Sq=Sk=32768, D=512`, passed with finite BF16 output
- `python benchmark/bench_flash_attn.py`: completed on BMG GPU 0
- `pre-commit run --files <PR changed files>`: passed
- full bundled FMHA/common_ops build completed with controlled parallelism (`SGL_KERNEL_BUILD_JOBS=20`)

### DIFF
--- a/src/sycl/kernels/flash_attention_v2/kernel/xe_fmha_fwd_kernel.hpp
+++ b/src/sycl/kernels/flash_attention_v2/kernel/xe_fmha_fwd_kernel.hpp
@@ -398,18 +398,18 @@ class XeFMHAFwdKernel {
       if constexpr (is_var_len) {
         auto qo_cumulative = s.seq_len_qo.cumulative_length;
         auto kv_cumulative = s.seq_len_kv.cumulative_length;
-        offset_q = s.num_heads_q * s.head_size_qk * qo_cumulative[idx_b];
+        offset_q = get<0>(p.dQ) * qo_cumulative[idx_b];
         // offset_k = s.num_heads_kv * s.head_size_qk * kv_cumulative[idx_b];
         // offset_v = s.num_heads_kv * s.head_size_vo * kv_cumulative[idx_b];
-        offset_o = s.num_heads_q * s.head_size_vo * qo_cumulative[idx_b];
+        offset_o = get<0>(p.dO) * qo_cumulative[idx_b];
         if (s.seq_len_kv_cache.cumulative_length) {
           auto kv_cumulative_cache = s.seq_len_kv_cache.cumulative_length;
           // Non-paged KV stores all batches in one contiguous ragged buffer, so each
           // batch starts at its cumulative KV offset. Paged KV uses the page table for
           // absolute addressing, so no base offset is applied here.
           if constexpr (!CollectiveMainloop::PagedKV) {
-            offset_k_cache = s.num_heads_kv * s.head_size_qk * kv_cumulative_cache[idx_b];
-            offset_v_cache = s.num_heads_kv * s.head_size_vo * kv_cumulative_cache[idx_b];
+            offset_k_cache = get<0>(p.dK_cache) * kv_cumulative_cache[idx_b];
+            offset_v_cache = get<1>(p.dV_cache) * kv_cumulative_cache[idx_b];
           }
         }
       }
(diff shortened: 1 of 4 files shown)


src/sycl/kernels/flash_attention_v2/xe_fmha_fwd_prefill_runner.hpp
```diff
@@ -357,40 +390,62 @@ struct PrefillRunner {
...
+        const size_t full_batch_workspace = FMHAPrefillKernel::get_workspace_size(one_batch);
+        if (full_batch_workspace <= cap_bytes) {
+          batch_slice = int(cute::max(size_t(1), cap_bytes / full_batch_workspace));
+          batch_slice = cute::min(batch_total, batch_slice);
+        } else {
+          const int q_tiles = cute::ceil_div(params.seqlen_q, int(get<0>(typename FMHAPrefillKernel::TileShapeQK{})));
+          const int target_heads = cute::ceil_div(cute::max(1, hw_info.sm_count), cute::max(1, q_tiles));
+          query_head_slice = cute::min(params.h, target_heads);
+          const int first_head_slice = get_query_head_slice_size(0, params.h, head_group_q, query_head_slice);
+          const size_t per_batch = FMHAPrefillKernel::get_workspace_size(
```
Valentine233: `query_head_slice` is deduced from `sm_count`. How to ensure `per_batch` does not exceed `cap_bytes`?

src/sycl/kernels/flash_attention_v2/kernel/xe_fmha_fwd_kernel.hpp
Valentine233, line 270: ScoreBlock2D only does slices on `batch` and `num_heads_q`. What if `seq_len_q * seq_len_kv` already exceeds `cap_mb`?

src/sycl/kernels/flash_attention_v2/xe_fmha_fwd_prefill_runner.hpp
```diff
@@ -357,40 +390,62 @@ struct PrefillRunner {
 
     const int batch_total = params.b;
     int batch_slice = batch_total;
+    int query_head_slice = params.h;
     if constexpr (CollectiveMainloop::ScoreBlock2D) {
       static const int cap_mb = [] {
         if (const char* env = std::getenv("FMHA_SCORE_WS_CAP_MB")) {
           return std::atoi(env);
         }
         return DefaultScoreWorkspaceCapMiB;
```
Valentine233, line 423: `DefaultScoreWorkspaceCapMiB` is hard-coded to `1024`. How is the value selected? OOM still happens if free memory is less than `1024`.

Valentine233 (changes requested):
In general, OOM still could happen with the current fix. Is it possible to disable ScoreBlock2D by default, or to enable ScoreBlock2D on a limited configs?

yuankuns:
> In general, OOM still could happen with the current fix. Is it possible to disable ScoreBlock2D by default, or to enable ScoreBlock2D on a limited configs?

In which case?

src/sycl/kernels/flash_attention_v2/xe_fmha_fwd_prefill_runner.hpp
yuankuns: query_head_slice is related to total amount of buffer. Now there is only 20 cores for bmg

src/sycl/kernels/flash_attention_v2/kernel/xe_fmha_fwd_kernel.hpp
yuankuns: There is q_tile in line 248

src/sycl/kernels/flash_attention_v2/xe_fmha_fwd_prefill_runner.hpp
yuankuns: There is a trade off between performance and memory usage. Given the fact that hd=512 exhaust slm/register already. For single kernel launch it will reach extreme long sequence to oom. In those cases, even non-scoreboard case can trigger oom

sunjiweiswift:
> > In general, OOM still could happen with the current fix. Is it possible to disable ScoreBlock2D by default, or to enable ScoreBlock2D on a limited configs?
> 
> In which case?

pls check this jira

tests/test_flash_attention.py
```diff
@@ -2245,6 +2246,144 @@ def test_flash_attn_varlen_cross_matrix(
...
+            # Non-paged HD512 uses 32 Q tiles per head. Q-tile chunking keeps
+            # one query head sufficient to cover BMG (20 Xe-cores) without a
+            # 4 GiB score buffer.
+            (1, 16, 2, 4096, 32768),
+            # A low-head, high-batch case must accumulate Q tiles across batch
+            # slices instead of allocating all eight score matrices at once.
+            (8, 1, 1, 4096, 32768),
+            # Exercises nested batch and GQA-head slicing: one unsliced batch
+            # needs 2 GiB and the complete problem needs 8 GiB.
+            (4, 8, 1, 4096, 32768),
```
Valentine233, line 2362: Could you please also check for these configs offline and make sure to pass, thanks!
```
(4, 16, 2, 32768, 32768)
(1, 32, 4, 32768, 32768)
(8, 8, 1, 32768, 32768)
(32, 4, 1, 32768, 32768)
(4, 8, 2, 32768, 32768)
```

airMeng:
replaced by #447

PR #422, opened 2026-08-25
Replace chunk prefill with AppendKV FMHA prefill

yuankuns:
Migrated from the former yuankuns/sgl-kernel-xpu branch.


yuankuns:
Closing this bulk migration PR.

PR #446, opened 2026-08-26
[MoE] Pick the W4A16 grouped-GEMM tile by rows-per-expert fill, not a threshold

airMeng:
Measured on Arc Pro B60, MXFP4 bf16 activations, gemm1 TFLOP/s (gemm2 tracks
    it). GPT-OSS TP=1/EP=1, 32 experts, N=5760, K=2880:

      avg_m |  64    96   128   129   160   192   256   512
      before| 48.5  49.3  49.2  24.8  34.8  47.0  66.0  66.5
      after | 60.0  49.4  66.4  38.1  53.2  64.1  67.4  68.1

### DIFF
--- a/src/GroupGemmW4A16Xe20.cmake
+++ b/src/GroupGemmW4A16Xe20.cmake
@@ -14,15 +14,18 @@ function(add_group_gemm_w4a16_xe20_inst POLICY ELEMENT_S ELEMENT_A SANITIZED)
     set(GROUP_GEMM_W4A16_XE20_INST_SRCS ${GROUP_GEMM_W4A16_XE20_INST_SRCS} PARENT_SCOPE)
 endfunction()
 
-# Policy menu (selected at runtime by average rows-per-expert):
-#   w4a16_policy_m_8   <_8,  _64,  _32>  — avg_m <= 4
-#   w4a16_policy_m_16  <_16, _64,  _32>  — avg_m <= 8
-#   w4a16_policy_m_32  <_32, _64,  _32>  — avg_m <= 128
-#   w4a16_policy       <_128,_256, _32>  — avg_m > 128
+# Policy menu. select_w4a16_tile_m() in GroupGemmW4A16Xe20.cpp picks one per
+# call from the average rows-per-expert:
+#   w4a16_policy_m_8         <_8,  _64,  _32>  — avg_m <= 4
+#   w4a16_policy_m_16        <_16, _64,  _32>  — avg_m <= 8
+#   w4a16_policy_m_32        <_32, _64,  _32>  — small avg_m
+#   w4a16_policy_m_64_n_128  <_64, _128, _32>  — mid avg_m
+#   w4a16_policy_m_128_n_128 <_128,_128, _32>  — large avg_m
 # group_size (32/64/128/256) is compiled into every unit as a runtime branch,
-# so it does not multiply the instance count. Total: 4 policies x 2 (int4/mxfp4)
-# x 2 (bf16/fp16 activation) = 16 units.
-foreach(policy w4a16_policy_m_8 w4a16_policy_m_16 w4a16_policy_m_32 w4a16_policy)
+# so it does not multiply the instance count. Total: 5 policies x 2 (int4/mxfp4)
+# x 2 (bf16/fp16 activation) = 20 units.
+foreach(policy w4a16_policy_m_8 w4a16_policy_m_16 w4a16_policy_m_32
+               w4a16_policy_m_64_n_128 w4a16_policy_m_128_n_128)
     foreach(act_tag bf16 fp16)
         if(act_tag STREQUAL "bf16")
             set(element_a "cutlass::bfloat16_t")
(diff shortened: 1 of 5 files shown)


src/sycl/GroupGemmW4A16Xe20.cpp
```diff
@@ -84,11 +84,55 @@ void w4a16_launch(
...
+// avg_m values that fill each tile exactly, in TFLOP/s. They are only ever
+// compared against each other, so the absolute scale does not matter -- what
+// matters is that a wider tile is worth more per row but wastes more of a
+// partial tile. GPT-OSS gemm1 (N=5760, K=2880) and DeepSeek-V4 gemm1
+// (N=4096, K=4096) agree on this ordering.
+constexpr int kW4A16TileM[] = {32, 64, 128};
+constexpr int kW4A16TileN[] = {64, 128, 128};
+constexpr float kW4A16TilePeakTflops[] = {49.0f, 64.0f, 68.0f};
+
+int select_w4a16_tile_m(int avg_m, int gemm_n) {
```
YangKai0616, line 119: The kernel uses a persistent grid. Changing the tile count also changes the number of resident WGs. When there aren’t enough tiles, excess WGs simply exit, leaving the machine idle, while the current model still reports 100% efficiency. 

The score only considers fill rate, not occupancy. Should we also account for XPU wave efficiency?

src/sycl/GroupGemmW4A16Xe20.cpp
```diff
@@ -263,17 +308,19 @@ SGL_KERNEL_EXPORT void moe_grouped_mm_nt_xe20_w4a16(
...
-      LAUNCH_W4A16(w4a16_policy_m_32); \
-    } else {                           \
-      LAUNCH_W4A16(w4a16_policy);      \
-    }                                  \
+#define DISPATCH_W4A16_POLICY()               \
+  do {                                        \
+    if (avg_m <= 4) {                         \
+      LAUNCH_W4A16(w4a16_policy_m_8);         \
+    } else if (avg_m <= 8) {                  \
+      LAUNCH_W4A16(w4a16_policy_m_16);        \
```
YangKai0616: Why didn’t the first two options enter `select_w4a16_tile_m`?

src/GroupGemmW4A16Xe20.cmake
```diff
@@ -14,15 +14,18 @@ function(add_group_gemm_w4a16_xe20_inst POLICY ELEMENT_S ELEMENT_A SANITIZED)
...
-#   w4a16_policy_m_16  <_16, _64,  _32>  — avg_m <= 8
-#   w4a16_policy_m_32  <_32, _64,  _32>  — avg_m <= 128
-#   w4a16_policy       <_128,_256, _32>  — avg_m > 128
+# Policy menu. select_w4a16_tile_m() in GroupGemmW4A16Xe20.cpp picks one per
+# call from the average rows-per-expert:
+#   w4a16_policy_m_8         <_8,  _64,  _32>  — avg_m <= 4
+#   w4a16_policy_m_16        <_16, _64,  _32>  — avg_m <= 8
+#   w4a16_policy_m_32        <_32, _64,  _32>  — small avg_m
+#   w4a16_policy_m_64_n_128  <_64, _128, _32>  — mid avg_m
+#   w4a16_policy_m_128_n_128 <_128,_128, _32>  — large avg_m
```
YangKai0616: Standardize to `m_X_n_Y`?

src/sycl/GroupGemmW4A16Xe20.cpp
YangKai0616, line 86: `m_X_n_Y` same here.

YangKai0616 (changes requested):
LGTM, with a few minor changes.

src/sycl/GroupGemmW4A16Xe20.cpp
airMeng: results of performance tuning

src/sycl/GroupGemmW4A16Xe20.cpp
airMeng: Occupancy is higher with the new policy actually

src/GroupGemmW4A16Xe20.cmake
airMeng: fixed in https://github.com/sgl-project/sgl-kernel-xpu/pull/446/changes/1f2983a8b797e89bb55d8ff4162e63bdac9ba607

src/sycl/GroupGemmW4A16Xe20.cpp
YangKai0616: To be precise, my point was about wave quantization, not occupancy.

For example: `waves = ceil(tiles / wgs)`, the score becomes `peak * fill_m * fill_n * tiles / (waves * wgs)`, which collapses to your current expression whenever the tile count is large enough. It only works with the low tile count (e.g. `EP=8` with few local experts).

But not worth blocking on, maybe need test and revisit it in a follow up tuning pass. This PR is good to go from my side!

YangKai0616 (approved):

PR #458, opened 2026-08-31
attention kvcache-mxfp4

sunjiweiswift:


airMeng:
I‘d suggest to drop the aggressive MX-FP4 KV cache feature on BMG and keep it for future HW only

PR #466, opened 2026-09-01
[Attn]Change minimum threshold for split KV

gaopengff:
With max_seq_len = 8192 and block_size = 128, **effective_blocks** is 64 and used 1 for num_split_kv. With this change it will use **num_xe_cores** for num_split_kv. This will improve perf for small batch decode with some models.
With following configs:
```python
q_seq_length = 1
kv_seq_length = 8192
head_dims = 128
num_q_heads, num_kv_heads = 8, 2
head_dim = 128
sliding widonw = False
page_size = 128
datatype = bf16
```
| batch | new time(ms) | old time (ms) | speed up |
|---|---|---|---|
| 1 | 0.0340105 | 0.122812 | 3.61x |
| 2 | 0.054323 | 0.095677 | 1.76x |
| 4 | 0.096719 | 0.124479 | 1.29x |
| 8 | 0.177396 | 0.236928 | 1.34x |


Valentine233:
Could you please provide the perf comparison with `python benchmark/bench_flash_attn.py`?

gaopengff:
> Could you please provide the perf comparison with `python benchmark/bench_flash_attn.py`?

Actually this specific is not in bench_flash_attn.py, do you want me to add it?

Valentine233:
> Actually this specific is not in bench_flash_attn.py, do you want me to add it?

No need to be added. Just want to make sure the modification won't cause regression on other cases.

gaopengff:
> > Actually this specific is not in bench_flash_attn.py, do you want me to add it?
> 
> No need to be added. Just want to make sure the modification won't cause regression on other cases.

I've run the benchmark locally and compared the data. There is no regression brough by this PR's change. Anyway, no test cases will reach changed code pass here. Here is detailed summary by AI:
1. 588/588 configs align exactly between old_attn.log and new_attn.log.
Median tflops difference is ~0.09%, mean ~0.29% — well within normal run-to-run measurement noise.
2. Only 2 rows differ by more than 5%, and both are extreme low-tflops/low-workload cases (row 185: batch=1, tiny local-window decode config, 0.016→0.019 TFLOPS; row 289: fp8_e4m3 page_size=128 case, 6.23→6.79 TFLOPS), where small absolute timing jitter translates into large relative swings.
3. No systematic regression or improvement pattern across configs (both faster/slower differences appear roughly symmetrically).

Valentine233 (approved):
Wait for the CI pass.

PR #467, opened 2026-09-01
[qknorm] Allow packed-QKV slices in XPU fused QK-Norm

YangKai0616:
## motivation

The current XPU JIT QK-Norm wrapper required `q`/`k` to be contiguous, which rejects the `q`/`k` slices of a fused QKV projection. This blocks callers with a packed QKV projection (e.g. `apply_qk_norm` in SGLang) from using the fused kernel instead of falling back to two out-of-place RMSNorms.

The kernel does not need this: it addresses a head as `base + token * stride(0) + head * head_dim` and takes the per-token row stride as a runtime parameter, so only the intra-token layout has to be fixed.

### DIFF
--- a/python/sgl_kernel/jit/norm.py
+++ b/python/sgl_kernel/jit/norm.py
@@ -166,6 +166,10 @@ class _XPUQKNormWrapper:
         import ctypes
 
         self._module = module
+        self._head_dim = head_dim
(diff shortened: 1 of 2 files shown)


airMeng:
@sunjiweiswift can you add the CI for JIT to avoid the next brokens?

YangKai0616:
Friendly ping @sunjiweiswift

airMeng (approved):

PR #469, opened 2026-09-02
perf(swiglu, norm): vectorized swiglu tail handling and register-cached RMSNorm path

jmunetong:
## Summary

Two changes to bandwidth-bound kernels. They are different kinds of change, and conflating them is the easiest way to misread the numbers below:

1. **swiglu** — a real optimization. The old "vectorized" kernel was not issuing vector memory traffic. It also read and wrote out of bounds on any shape whose element count wasn't a multiple of 4.
2. **norm** — a deletion. `RMSNorm.cpp` contained two complete RMS-norm implementations; `rmsnorm` and `fused_add_rmsnorm` were wired to the slower one. Rerouting them onto the faster one and deleting the loser is the entire 1.31–1.38×. No kernel became faster.

---

### Fixes an out-of-bounds read and write

The old kernel's guard was block-granular:

```cpp
if (base >= total_pairs) return;   // then unconditionally touches base..base+3
```

A work-item with `base < total_pairs <= base + 3` cleared the guard and then read `x` and **wrote `y` up to 3 elements past the end** — elements a second, separately launched scalar kernel also wrote. Triggered whenever `B*H % 4 != 0` and the vector work-item count wasn't a multiple of 256.

The new kernel bounds-checks per element in the single work-item past the vector range, so the op is also now **one launch instead of two**.

### Makes the vectorization real

`x` is `[B, 2H]` with gate/up interleaved (gate at `2i`, up at `2i+1`); `y` is `[B, H]`. The old kernel used `sycl::vec<float,4>` as a register type but addressed memory one element at a time at stride 2 (`x[2 * (base + i)]`) with scalar stores — a gather and a scatter, not a block transfer.

`SwigluInterleavedVecKernel<scalar_t, VP>` (VP = 8 for bf16/fp16, 4 for fp32) instead gives each work-item **one 32 B contiguous block load** of the gate/up pair-block, de-interleaves in registers, and does **one 16 B contiguous block store**.

The pair-block is the only formulation that works: gate lives only at even indices, so a "gate-only vector load" is a stride-2 gather by construction. There is no contiguous gate slice to load.

Also here: the activation is factored into one `swiglu_gpt_oss_elem` helper documented as needing to stay expression-for-expression identical to the fused BF16 grouped-GEMM epilogue (`kernels/moe/xe20/common/activation.hpp`); `sycl::native::exp` for 2-byte dtypes only (fp32 keeps precise `exp`, where 1e-6 relative error would be ~10 ULP of the result); and an alignment guard falling back to the scalar kernel, newly required because the vector path `reinterpret_cast`s pointers.

---

## 2. norm (`src/sycl/RMSNorm.cpp`, net −436 lines)

Two families existed. Only the Gemma entry points used the good one:

| | old "rstd" family | "no-rstd" family |
|---|---|---|
| `x` between the two passes | **re-read from DRAM** | **cached in registers** |
| `rstd` | written to a global `at::empty({M})`, re-read | kept in a register |
| used by | `rmsnorm`, `fused_add_rmsnorm` | `gemma_rmsnorm`, `gemma_fused_add_rmsnorm` |

This PR points all four entry points at the no-rstd family and deletes the rstd family. The `rstd` tensor was internal scratch, never returned to the caller; dropping that per-call allocation is a host-side win the device-time numbers cannot see.

To share the residual-add path between the llama-style and Gemma-style norms, `AddNoRstdForward` is a mixin over a base norm that supplies `apply_weight` — Gemma hides it with the `(1 + gamma)` variant, since it stores `gamma - 1`.

**One invariant to review:** `reduce_combine` writes the residual sum to `add_data`, never back to `X_data`, so the **non-cached** `update()` must re-read from `add_data` — `X_data` still holds the un-summed `x`. That path is untested (see gaps).

---

## Results

A/B on BMG, `ZE_AFFINITY_MASK=4,5,6,7`. Baseline `cf20cbd`, whose three touched files are byte-identical to `origin/main`. Device time via profiler, 110 configs per arm, DRAM ceiling measured at 395.7 (baseline) / 394.6 (candidate) GB/s — 0.27% apart. Scope: the **58 configs bandwidth-demonstrable in both arms**, plus 1 re-measured.¹ Per-config spreads ≤0.9%.

| op | n | min | median | max |
|---|---|---|---|---|
| `fused_add_rmsnorm` | 13 | 1.036 | **1.382×** | 1.933 |
| `swiglu` | 7 | 0.983 | **1.441×** | 1.450 |
| `rmsnorm` | 13 | 1.138 | **1.311×** | 1.591 |
| `gemma_rmsnorm` | 15 | 0.998 | 1.001× | 1.007 |
| `gemma_fused_add_rmsnorm` | 11 | 0.998 | 1.001× | 1.007 |

**swiglu**: bf16/fp16 go from 68–69% to **99% of the DRAM ceiling**. fp32 was already at 99% and is flat.

**norm — the numbers are what the roofline predicts.** The deleted path read `x` from DRAM twice:

- `rmsnorm`: 3 DRAM touches → 2. Bound **1.5×**, measured 1.311× median.
- `fused_add_rmsnorm`: 6 → 4. Bound **1.5×**, measured 1.382× median. The 1.933× max exceeds that bound, so something shape-specific also helps there; not claimed as explained.

**The gemma rows are not an untouched control.** `GemmaRMSNormNoRstdForward` was itself rerouted through the new mixin, and its old symbol is absent from the new binary (2304 → 0). Their flatness (n=35, median **+0.03%**, max 0.75%) is therefore evidence that **the shared-mixin refactor is performance-neutral** — a different and more useful claim than "we didn't touch it."

**The intended kernels ran**, confirmed at the symbol level: `…RMSNormForwardELb1E…` 384 → 0, `FusedNormKernelFunctor` absent entirely, `SwigluVec4Kernel` → `SwigluInterleavedVecKernel<…,8>`.

---

## Tests

`tests/test_norm.py` 638 passed, `tests/test_swiglu_with_alpha_limit.py` 977 passed. The in-run numerics gate passes 110/110 in both arms with an identical worst-case `max_rel` of 0.0078125, so the reroute does not shift the tolerance class. Tolerances were only tightened (fp32 `fused_add` 1e-3 → 1e-4), never loosened.

Swiglu tail coverage is thorough: every nonzero remainder mod 8 and mod 4, `B*H = 1`, no-tail controls, all three dtypes, at 1 output ULP against an fp32-narrowed reference — so a one-element tail-bound error fails.

**Known gaps, being addressed:**

- The non-cached `update()` fallback (`Plane/update_vec_size > 4096`) is reached by no test and no benchmark config, and it is where the new `AddNoRstdForward` residual-source invariant lives. Highest-risk gap in the PR.
- `rmsnorm` has no fp32 correctness test, though fp32 is benchmarked and is a distinct instantiation.
- The odd-`N` shapes this PR reports numbers for are absent from correctness tests: every vec-1 instantiation with `ITERS >= 2` is untested.
- No tight-tolerance case where `limit` actually binds, so the gate/up clamp asymmetry is unpinned at ULP precision; `native::exp` is unverified beyond `|x| ≈ 5`.

---

## Not fixed here

Odd-`N` shapes remain far off the roofline: `tok8193_h2881` bf16 `rmsnorm` improves 22% → 29% of ceiling, still **3.4× off the bound** (114.8 vs 394.6 GB/s). The exact tail bound made that shape *correct*, not *fast*. Pre-existing and a property of the no-rstd kernel, not introduced here — `gemma_rmsnorm` sits at the same 29% in *both* arms. Remaining headroom ≈ 822 → 239 µs; follow-up.

No end-to-end measurement. The wins are prefill-shaped; the decode-shaped configs are precisely the submission-bound ones that will not translate.

<sub>¹ `rmsnorm tok8192_h4096` bf16 was unstable in the baseline main pass (9 samples at ~448.6 µs, then excursions to 535/505/481, 19.4% spread) and is excluded from it. Its 1.309× ratio comes from the baseline's independent late drift-canary pass, which read 449.23 µs at 0.14% spread, agreeing with the 9 good samples. Main-pass-only, `rmsnorm` p50 is 1.324×, so the reported 1.311× is the more conservative figure.</sub>


jmunetong:
@airMeng could you review this when you have a chance? I will be running E2E model profiling today/tomorrow and will be sharing the results here.

tests/test_swiglu_with_alpha_limit.py
```diff
@@ -47,5 +68,112 @@ def test_swiglu_gpt_oss_sigmoid_alpha(batch_size, hidden_size, alpha, limit, dty
...
+        neighbour, torch.full_like(neighbour, sentinel)
+    ), f"kernel wrote past the end of y (dtype={dtype}, B*H={total_pairs})"
+    torch.testing.assert_close(
+        second,
+        swiglu_gpt_oss_sigmoid_alpha_ref_fp32(x, alpha, limit),
+        rtol=TIGHT_TOL[dtype][0],
+        atol=TIGHT_TOL[dtype][1],
+    )
+
+
```
gaopengff, line 177: Since torch.randn() will generate values most range [-3, 3], the clamp operation was not tested in most cases as the limit is larger(7.0). Suggest add a new test case with larger input and smaller limit value. Like the following:
```python
@pytest.mark.parametrize("dtype", [torch.float32, torch.bfloat16, torch.float16])
def test_swiglu_clamp_asymmetry(dtype):
    """limit small relative to input scale: forces gate's upper-only clamp
    and up's symmetric clamp to both bind, on both positive and negative
    inputs, so a clamp-shape bug (e.g. accidentally symmetric gate clamp)
    is caught at ULP precision."""
    alpha, limit = 1.702, 0.1  # small limit vs input scale
    hidden = 128
    x = torch.randn((256, hidden * 2), dtype=dtype, device="xpu") * 5.0  # scale up so |x| >> limit

    output = swiglu_gpt_oss_sigmoid_alpha(x, alpha, limit)
    output_ref = swiglu_gpt_oss_sigmoid_alpha_ref_fp32(x, alpha, limit)

    rtol, atol = TIGHT_TOL[dtype]
    torch.testing.assert_close(output, output_ref, rtol=rtol, atol=atol)
```

src/sycl/RMSNorm.cpp
```diff
@@ -63,226 +60,6 @@ static inline RowStrides get_row_strides(const Tensor& t) {
   return {outer_stride, inner_size, inner_stride};
 }
 
-template <typename scalar_t, typename weight_t, typename mean_t = float>
-class RMSNormForward : public NormForward<scalar_t, weight_t, true> {
```
CaoE, line 67: Since you're not using the older version, do we still need the parts of the header file that aren't being used ?

jmunetong: You are correct. We don't need this anymore. I have addressed them now

jmunetong:
### Model-level effect on GPT-OSS-120B

Intel Arc Pro B60, TP=4, MXFP4 W4A16, `bench_serving` random 4096-in / 1024-out, `--max-concurrency 1`, graph capture off, **n=1 per revision**. Without/with = **v2 `965f176` → v3 `b53efc5`** (this branch).

**Prefill, device self-time**

| Bucket | Kernel | Without | With | Δ | ÷ projection |
|---|---|---:|---:|---:|---|
| MoE swiglu | `SwigluVec4Kernel` → `SwigluInterleavedVecKernel` | 9.427 ms | 6.329 ms | **−3.098 (−32.86%)** | 1.52x → **1.02x** |
| Norms | `FusedNormKernelFunctor` → `RMSNormNoRstdKernelFunctor` | 19.769 ms | 14.950 ms | **−4.819 (−24.38%)** | → **0.996x** |
| **Sum** | | **29.196 ms** | **21.279 ms** | **−7.917 (−27.12%)** | |

**Decode, device self-time**

| Bucket | Without | With | Δ | |
|---|---:|---:|---:|---|
| Norms | 0.432 ms/step | 0.176 ms/step | **−0.256 (−59.26%)** | real |
| MoE swiglu | 0.113 ms/step | 0.105 ms/step | −0.008 (−7.08%) | **not quotable** — launch-bound at batch=1 |

**Scope of the two tables above**

| Item | Value |
|---|---|
| Share of the branch-attributable prefill win | −7.917 of −9.651 ms = **82.0%** |
| Share of the comms-excluded prefill total | **−3.59%** of 220.466 ms |
| E2E ceiling, prefill ×1 + decode ×1023 steps | **−277.99 ms = −0.40%** — arithmetic from kernel self-time, **not a measured A/B of this PR** |
| Measured branch-level E2E, same step | **+2,307.49 ms (+3.36%)** — wrong direction; that step also carries a oneCCL runtime swap, not this branch |
| Same two kernels two campaigns later | swiglu 6.354 ms (**1.023x**), norms 15.047 ms (**0.996x**) — win did not decay |
| Attribution basis | kernel symbol changed in both buckets; the 3 other changes in that step touch other buckets — **not a single-variable A/B** |
| Norms ÷ projection, "without" cell | withdrawn: v3's projection priced an *unfused* RMSNorm against a kernel this PR had already fused, so the old 2.62x → 1.98x pair was ≈ the buffer-count ratio |
| 20b (TP=1) companion | prefill **−10.270 ms** |

tests/test_swiglu_with_alpha_limit.py
jmunetong: Fixed

gaopengff (approved):
LGTM

CaoE (approved):

PR #477, opened 2026-09-08
[CRI] upstreaming CRI specific kernels and changes

mkumargarg:
fix the regression in running the test for CRI. Also using CPU reference for CRI for some tests.


src/sycl/kernels/moe/xe20/w4a16/mxfp4_dequant.hpp
```diff
@@ -146,7 +159,8 @@ mxfp4_reorder_folded(intel::uchar4 const& src0, intel::ushort8& dst0, intel::vec
         : "rw"(src0), "rw.u"(shifts), "rw"(muls));
   }
 #elif defined(__SYCL_DEVICE_ONLY__) && defined(SYCL_INTEL_TARGET)
-#error "MXFP4 folded reorder is only supported on SYCL_INTEL_TARGET 20 (BMG/Xe20)"
+  static_assert(
+      mxfp4_folded_unsupported<Kind>, "MXFP4 folded reorder is only supported on SYCL_INTEL_TARGET 20 (BMG/Xe20)");
```
Xia-Weiwen: This part won't be built for Xe3 anymore with this patch: https://github.com/sgl-project/sgl-kernel-xpu/pull/489

tests/utils.py
```diff
@@ -16,6 +18,20 @@ def get_device():
...
+    if not torch.xpu.is_available():
+        return False
+    from sgl_kernel.utils import is_xe2_arch
+
+    return is_xe2_arch()
+
+
+def get_reference_device():
+    """Reference device for eager PyTorch baselines."""
+    return get_device() if is_xe2_device() else torch.device("cpu")
```
kareemshaik80, line 32: we still need this condition? we can make it always device instead of cpu?

kareemshaik80 (approved):
LGTM

PR #481, opened 2026-09-09
[XPU] flash_mla_decode: derive latent/rope dims instead of hardcoding…

desmond-intel:
The flash_mla_decode wrapper pinned D_latent=512 and D_rope=64 and asserted the query dims equal them, so an MLA model whose latent/rope dims differ from DeepSeek's (e.g. MiniCPM3-4B: 256/32) tripped the assert before reaching the kernel. The SYCL kernel is dimension-generic -- it tiles the latent dim and predicates the tail, and flash_mla_prefill in this file already derives its dims the same way -- so 512/64 is one instance, not a constraint.

Derive D_latent/D_rope from the q tensors and keep only the D_ckv == D_latent + D_rope check. D_latent still sizes the output, so it must be the real latent width rather than a hardcoded 512.

### DIFF
--- a/python/sgl_kernel/attention.py
+++ b/python/sgl_kernel/attention.py
@@ -74,11 +74,11 @@ def flash_mla_decode(
 
     _, PAGE_SIZE, D_ckv = kv_c_and_k_pe_cache.shape
 
-    D_latent = 512
-    D_rope = 64
-    assert D_q_nope == D_latent
-    assert D_q_pe == D_rope
-    assert D_ckv == D_latent + D_rope
+    D_latent = D_q_nope
+    D_rope = D_q_pe
+    assert (
+        D_ckv == D_latent + D_rope
+    ), f"kv dim {D_ckv} must equal D_latent({D_latent}) + D_rope({D_rope})"
 
     MAX_HEADS = 128
     assert H <= MAX_HEADS, f"H must be <= {MAX_HEADS}, but got {H}"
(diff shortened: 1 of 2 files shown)


polisettyvarma (approved):

python/sgl_kernel/attention.py
```diff
@@ -74,11 +74,13 @@ def flash_mla_decode(
 
     _, PAGE_SIZE, D_ckv = kv_c_and_k_pe_cache.shape
 
-    D_latent = 512
-    D_rope = 64
-    assert D_q_nope == D_latent
-    assert D_q_pe == D_rope
-    assert D_ckv == D_latent + D_rope
+    # Derive dims from the tensors (like flash_mla_prefill); the kernel tiles the
+    # latent and predicates the tail, so 512/64 is not a hard extent.
```
pralay-das: remove this commented lines.

pralay-das (commented):
also add a test case for which dims currently you are fixing the issue in the `tests/test_flash_mla_decode.py` file.

python/sgl_kernel/attention.py
desmond-intel: Done, removed the comment lines.

python/sgl_kernel/attention.py
```diff
@@ -74,11 +74,13 @@ def flash_mla_decode(
...
+    D_latent = D_q_nope
+    D_rope = D_q_pe
+    assert (
+        D_ckv == D_latent + D_rope
+    ), f"kv dim {D_ckv} must equal D_latent({D_latent}) + D_rope({D_rope})"
```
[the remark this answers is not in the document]
desmond-intel: Added a 256/32 parametrization; passes on XPU.

desmond-intel:
@pralay-das added a 256/32 test case in tests/test_flash_mla_decode.py (parametrized alongside 512/64); passes on XPU.

pralay-das (approved):

PR #486, opened 2026-09-11
Add AOT SYCL Mamba HiCache transfer kernels for XPU

kpjeeja:
## Motivation

Mamba/hybrid models keep conv+ssm state in a paged GPU cache and back it up to a
layer-major host cache. On XPU that move currently falls back to PyTorch
advanced indexing, which is a per-element gather/scatter: it flat-lines at
~32-38 GB/s on the backup path regardless of transfer size, because it is bound
by indexing overhead rather than bandwidth. This adds the two AOT SYCL kernels
that CUDA already has (`csrc/kvcacheio/transfer_mamba.cuh`,
`TransferMambaKernel::run_pf_lf` / `::run_lf_pf`).

## What this adds

```
load  (pf->lf): dst[dst_idx[i]]        <- src[src_idx[i], layer_id]
backup(lf->pf): dst[dst_idx[i], layer] <- src[layer, src_idx[i]]  (all layers)
```

- `transfer_kv_mamba_pf_lf` moves a single layer.
- `transfer_kv_mamba_lf_pf` moves every layer in one launch.

`item_size` and `layout_dim` are taken in **BYTES**, matching the CUDA kernels.

## Implementation notes

**Runtime copy width, so it is dtype-agnostic.** The copy unit (8/4/2/1 B) is
picked at runtime from the OR of the two pointer addresses and the byte strides,
so one kernel serves every state dtype and no per-dtype instantiation is needed.
Well-aligned shapes move 8 B per work-item; an unaligned `item_size` degrades to
a 1 B unit instead of failing. This is a deliberate difference from the KV
kernels above it in the same file, which hard-require `item_size % 8 == 0`.

**Chunked grid.** The natural mapping is one work-group per copy unit (item for
load, item x layer for backup). With few, large items that leaves the device
idle - 16 pages of 64Ki elements is 16 work-groups on a 160-CU GPU, which
measured *slower* than the PyTorch indexed copy - so each unit is split into
contiguous chunks until the grid fills the device. With enough units to fill it
(the common HiCache case) the split collapses back to one group per unit.

**N-D ranges rather than a flat range.** Expressed as `nd_range<2>` (load) and
`nd_range<3>` (backup) so the hardware supplies the item/layer/chunk indices.
Recovering them from a flat group id costs an int64 div+mod per work-item, which
measurably slowed the small-item shapes.

**Backup takes a tensor, not a pointer array.** Where CUDA passes a uint64 array
of per-layer pointers, this takes the contiguous layer-first tensor plus its
per-layer byte stride: dereferencing a host-built pointer array is not portable
across SYCL runtimes. Numerically identical.

## Tests

`tests/test_transfer_mamba.py` adds **91 cases**. These kernels are pure indexed
block copies, so a correct result is bit-exact and every comparison uses
`rtol=0, atol=0`:

- 4 dtypes x 4 item sizes x 3 layer counts
- the multi-chunk grid path (few very large items, with one size deliberately
  not a multiple of the chunk stride so the tail chunk is short)
- an unaligned item size that forces the 1-byte copy unit
- empty indices
- a load/backup roundtrip

**Result: 91/91 passed** on BMG.

## Benchmark

`benchmark/bench_transfer_mamba.py`, BMG, `dtype=bfloat16`, `num_layers=4`.
`torch` = advanced-indexing gather/scatter, `sycl` = these kernels. Bandwidth
counts read+write (2x payload).

| op | items | elems | MiB | torch us | sycl us | speedup | torch GB/s | sycl GB/s |
|---|---|---|---|---|---|---|---|---|
| load | 16 | 1024 | 0.06 | 10.26 | 5.00 | 2.05x | 6.4 | 13.1 |
| backup | 16 | 1024 | 0.25 | 18.07 | 5.16 | 3.50x | 14.5 | 50.8 |
| load | 16 | 8192 | 0.50 | 17.08 | 7.60 | 2.25x | 30.7 | 68.9 |
| backup | 16 | 8192 | 2.00 | 66.20 | 8.07 | 8.20x | 31.7 | 259.8 |
| load | 16 | 65536 | 4.00 | 63.10 | 11.56 | 5.46x | 66.5 | 362.8 |
| backup | 16 | 65536 | 16.00 | 450.78 | 29.64 | 15.21x | 37.2 | 566.1 |
| load | 128 | 1024 | 0.50 | 15.68 | 6.20 | 2.53x | 33.4 | 84.6 |
| backup | 128 | 1024 | 2.00 | 65.16 | 9.64 | 6.76x | 32.2 | 217.7 |
| load | 128 | 8192 | 4.00 | 63.70 | 12.45 | 5.12x | 65.8 | 336.9 |
| backup | 128 | 8192 | 16.00 | 450.36 | 29.43 | **15.30x** | 37.3 | **570.1** |
| load | 128 | 65536 | 32.00 | 465.31 | 62.24 | 7.48x | 72.1 | 539.1 |
| backup | 128 | 65536 | 128.00 | 3614.11 | 323.70 | 11.16x | 37.1 | 414.6 |
| load | 512 | 1024 | 2.00 | 36.41 | 10.36 | 3.51x | 57.6 | 202.3 |
| backup | 512 | 1024 | 8.00 | 229.56 | 24.53 | 9.36x | 36.5 | 341.9 |
| load | 512 | 8192 | 16.00 | 229.27 | 33.28 | 6.89x | 73.2 | 504.1 |
| backup | 512 | 8192 | 64.00 | 1818.07 | 155.36 | 11.70x | 36.9 | 431.9 |
| load | 512 | 65536 | 128.00 | 1994.09 | 312.08 | 6.39x | 67.3 | 430.1 |
| backup | 512 | 65536 | 512.00 | 14322.00 | 1354.14 | 10.58x | 37.5 | 396.5 |

**Aggregate over all 18 configs: geomean 6.26x** (range 2.05x - 15.30x)

- `load` (n=9): geomean **4.16x**, 13-539 GB/s vs torch 6-73 GB/s
- `backup` (n=9): geomean **9.42x**, 51-570 GB/s vs torch 15-38 GB/s

Reading of the data:

- **Backup is where this pays.** torch sits at 31.7-37.5 GB/s on all nine backup
  configs across a 2000x size range (0.25 MiB to 512 MiB) - that is per-element
  scatter overhead, not bandwidth. The SYCL kernel scales with the transfer
  instead, peaking at 570 GB/s, which is why backup speedup is roughly 2x the
  load figure.
- **Peak is mid-sweep, not at the largest shape.** Best bandwidth is at a 16 MiB
  working set; it declines to ~396 GB/s at 512 MiB where the transfer is
  DRAM-bound with no cache left to exploit.
- **The 1024-element shapes are launch-latency-bound, not bandwidth-bound.** The
  SYCL floor is ~5-10 us there regardless of layout. `load / 16 / 1024` moves
  only 64 KiB, so its 13.1 GB/s is fixed overhead - note torch is worse still on
  that shape at 6.4 GB/s, the lowest number in the table. Not expected to matter
  in a real HiCache path.

### DIFF
--- a/benchmark/bench_transfer_mamba.py
+++ b/benchmark/bench_transfer_mamba.py
@@ -1,0 +1,169 @@
+"""Benchmark the Mamba HiCache transfer kernels: PyTorch vs AOT SYCL.
+
+Both providers move exactly the same bytes -- the kernels are pure gather /
+scatter copies with no arithmetic -- so the interesting number is achieved
+bandwidth:
+
+  torch : PyTorch advanced-indexing copy (what runs without a kernel)
+  sycl  : sgl_kernel.transfer_kv_mamba_*  (AOT SYCL, compiled into the wheel)
+
+  load   : dst[dst_idx[i]]        <- src[src_idx[i], layer_id]
+  backup : dst[dst_idx[i], layer] <- src[layer, src_idx[i]]   (all layers)
+"""
+
+import itertools
+
+import pandas as pd
+import torch
+import triton
+
+try:
+    from sgl_kernel import transfer_kv_mamba_lf_pf, transfer_kv_mamba_pf_lf
+
+    HAS_SYCL = True
+except ImportError:
+    HAS_SYCL = False
+    print("Warning: sgl_kernel Mamba transfer kernels not available")
+
+all_results = []
+
+# Layers copied per backup call (a real Mamba HiCache backup is all-layer).
+NUM_LAYERS = 4
+DTYPE = torch.bfloat16
+
+
+def _make_indices(pool, num_items, device):
+    src_idx = torch.randperm(pool, device=device)[:num_items].to(torch.int64)
+    dst_idx = torch.randperm(pool, device=device)[:num_items].to(torch.int64)
+    return src_idx, dst_idx
+
+
+def torch_load_pf_lf(src, dst, src_idx, dst_idx, layer_id, item_elems, num_layers):
+    """PyTorch reference gather: page_first -> single-layer."""
+    dst.view(-1, item_elems)[dst_idx] = src.view(-1, num_layers, item_elems)[
+        src_idx, layer_id
+    ]
+
+
+def torch_backup_lf_pf(src_layers, dst, src_idx, dst_idx, item_elems, num_layers):
+    """PyTorch reference scatter: layer_first -> page_first (all layers)."""
+    s = src_layers.view(num_layers, -1, item_elems)
+    dst.view(-1, num_layers, item_elems)[dst_idx] = s[:, src_idx, :].permute(1, 0, 2)
+
+
+def bytes_moved(op, num_items, item_elems, num_layers, elem_size):
+    """Bytes read + written by one call (copy => 2x the payload)."""
+    items = num_items if op == "load" else num_items * num_layers
+    return 2 * items * item_elems * elem_size
+
+
+# (num_items, item_elems, op). item_elems spans a small per-page Mamba state up
+# to a large fused conv+ssm state; num_items spans a short prefix-cache hit up to
+# a full-batch backup.
+configs = list(
+    itertools.product(
+        [16, 128, 512],  # num_items (pages transferred)
+        [1024, 8192, 65536],  # item_elems (state elements per page per layer)
+        ["load", "backup"],
+    )
+)
+
+
+@triton.testing.perf_report(
+    triton.testing.Benchmark(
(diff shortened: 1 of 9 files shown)


kpjeeja:
@sspintel @kareemshaik80 can you please help review this PR?

kareemshaik80 (commented):
LGTM!

kareemshaik80:
Enable CI.

sspintel (approved):
LGTM

PR #492, opened 2026-09-16
ci(xpu): add a CRI (Xe3P) build gate to PR CI

arathi-hlab:
## Motivation

A PR that compiles on BMG can still break Xe3P (CRI), and today nothing catches that before it lands. This adds a build-only `build-cri` gate to `pr-test-xpu.yml`.

CRI is pre-silicon, so no kernel runs and nothing is packaged or installed. It runs on `sglang-cri`, behind the same `run-ci` label gate as `build-and-test`, and builds `refs/pull/N/merge` — the PR on top of current `main`. The existing BMG flow is untouched (`build-and-test` is byte-identical).

## Modifications

Toolchain comes from `intel/deep-learning-essentials:2026.0.0-devel-ubuntu24.04` — the image `Dockerfile.xpu_kernel` itself `FROM`s — rather than the BMG nightly `intel/sgl-kernel-xpu-dev:latest`, which is confusing to pull on a CRI runner and whose prebuilt tree is a BMG `pip install .` this job ignores. On top of that base only `ocloc` plus its IGC/GMM dependencies are installed (pinned to the same versions as `Dockerfile.xpu_kernel`), because oneAPI does not ship `ocloc` and the Xe3P AOT device link needs it. `libze1`/`opencl-icd` and the media drivers are skipped: nothing executes here, and `DPCPP_SYCL_TARGET` is explicit so Level Zero device detection never runs.

Notes on the build step:

* Direct CMake, not pip: the gate must produce no artifact. A verify step asserts no wheel exists so a future edit reintroducing pip is caught.
* `DPCPP_SYCL_TARGET=cri` is passed explicitly. `BUILD_TARGET_DEVICE` is only consulted when Level Zero detection fails, so auto-detect would silently emit Xe2 on a BMG-equipped runner.
* `SKBUILD_SABI_*` and `__INTEL_LLVM_COMPILER` are supplied by hand to replace what scikit-build-core and the toolkit probe would set, making the configure flag-for-flag identical to a packaged build.
* `CC`/`CXX` are deliberately not set: `BuildFlags.cmake` gates the whole SYCL flag block on a GNU host compiler.
* `icpx` is put on `PATH` instead of sourcing `setvars.sh`, which would shadow the libsycl torch bundles and break `import torch`.
* Parallelism is derived from RAM rather than core count, since the compiler launcher kills the build on OOM.

The verify step greps the configure log and `CMakeCache` for the Xe3P markers, rejects a `bmg` fallback, and requires that `.so` files were produced, so a silently-BMG or configure-only run cannot report success.

## Checklist

- [x] Build-only: no kernels executed, no wheel produced.
- [x] Existing BMG `build-and-test` job unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pralay-das:
hi @arathi-hlab, could you please check this PR? it is pending from last week. Without the CI there are CRI specific builds and tests are failing which are very difficult to track.

cc. @mkumargarg

PR #496, opened 2026-09-17
perf(moe): fuse the W4A16 MoE prepare path at M==1 and memoize the bf16->fp32 bias promotion

jmunetong:
## Motivation

Two per-forward costs on the MXFP4 W4A16 MoE decode path are pure overhead, and both are visible as kernel launches in a gpt-oss-120b TP=4 decode trace:

1. **The fp32 bias promotion ran on every forward.** The Xe2 grouped GEMMs take the bias as `[E, N] float32` (`src/sycl/GroupGemmW4A16Xe20.cpp:195`, enforced at `:269-271`), but `w13_weight_bias` / `w2_weight_bias` arrive bf16 and are never promoted on XPU, so the cast was re-done per call — 2 launches per `fused_experts`, **72 of the 73 `UnrolledElementwiseKernel<CopyScalarFunc<c10::BFloat16>, ..., LoadWithCast<1>, StoreWithCast<1>>` launches/step** at 36 layers. These are *weight* casts: the cost does not shrink with batch size.
2. **MoE bookkeeping ran four kernels where one suffices.** `prepare_moe_input_small` already does the per-expert histogram, exclusive scan, arg-sort and token scatter in one kernel, but was gated on `use_fp8_weight`. The three outputs it does not produce (`problem_sizes1`, `problem_sizes2`, `a_map`) are dead on the W4A16 path — the grouped GEMMs take `expert_offsets` plus the shuffled activations, and the combine takes `c_map` — so the substitution is legal there too.

Measured effect of (2) in-trace: **4 bookkeeping launches/layer → 1, i.e. 144 → 36 launches/step, 0.3111 → 0.1207 ms/step device self-time = −0.190 ms/step.** That was a falsifiable launch-count prediction made before the run, and it held.

The fused kernel is a **single work-group launch** (`MoEPrepareInputs.cpp:742`, global range == local range), so its scatter cost grows with `num_tokens * TopK * hidden_dims` while its grid does not. It wins only at one token, via the kernel's own `input_rows_ == 1` fast path. Hence the gate is the exact equality `M == 1`, not a tuned threshold:

| `num_tokens` | 2 | 4 | 8 | 14 |
|---|---:|---:|---:|---:|
| fused vs four-kernel, gate opened to all bf16 contiguous W4A16 | **1.020x slower** | **1.036x** | **1.047x** | **1.070x** |

Monotone, and far outside the ~0.3 µs within-arm spread. That measurement *is* the justification for the gate.

## Modifications

| Commit | Change | Files |
|---|---|---|
| `7460aac` | Memoize the bf16→fp32 MoE bias promotion out of the hot path (`_bias_as_float32`), plus the microbench harness. | `python/sgl_kernel/moe.py`, `benchmark/bench_moe_dispatch_casts.py` |
| `c8f1f4a` | Add 7 small-m N-width policies (ids 7–13) for the W4A16 grouped GEMM, the `SGL_W4A16_POLICY_ID` override, and the sweep harness. **Unreachable by default.** | `src/sycl/kernels/moe/xe20/w4a16_launch_policy.hpp`, `src/sycl/GroupGemmW4A16Xe20.cpp`, `src/GroupGemmW4A16Xe20.cmake`, `benchmark/bench_moe_w4a16_policy_sweep.py` |
| `35e368b` | Use the fused `prepare_moe_input_small` kernel on the W4A16 path, gated to `M == 1`. `fused_experts`: 8 kernel launches → 5. Bit-exact. | `python/sgl_kernel/moe.py` |
| `1feaa6d` | Correct a false comment claiming the subgroup-tile *count* was invariant across N widths. | `src/sycl/kernels/moe/xe20/w4a16_launch_policy.hpp` |

Total: 6 files, +1262 / −10.

**`7460aac` — why the cache holds a strong reference, deliberately.** The cache is keyed on `id(bias)` and each entry holds a **strong** reference to the source tensor. That strong ref is what makes `id()` sound: a Python id can only be recycled after the object dies, and pinning the source alive makes recycling impossible. `data_ptr()` was rejected as a key precisely because the **XPU caching allocator recycles addresses**, so a freed-then-reallocated tensor could collide with a stale entry and be served the wrong promotion silently. The retained objects are *static weights* that live for the process anyway, so this is not a leak — it is a correctness requirement of the key. The cache is bounded (`_MOE_BIAS_F32_CACHE_MAX = 256`) so a caller churning short-lived bias tensors cannot grow it without limit, and a `_version` guard invalidates on in-place mutation. The version read goes through `_version_or_none` because tensors created under `torch.inference_mode()` *raise* on `_version` access and sglang loads weights under exactly that.

**`c8f1f4a` — measurement infrastructure, present and disabled on purpose.** The 7 new policies are reachable **only** via the debug env var `SGL_W4A16_POLICY_ID`; `select_w4a16_policy_id()` still returns 0–6, so **no layer changes behaviour by default**. It was built to test the hypothesis that decode's 8x64 tile leaves parallelism on the table, and **that hypothesis was refuted**: every small-m tile has `SG_N = 16`, so all of them decompose gemm1 into 368 and gemm2 into 720 16-wide subgroup tiles and fill the same 0.57/1.12 waves of the 640-subgroup persistent grid. Widening N while keeping the mainloop barrier **costs** 2–3% (8x128 0.97x, 8x256 0.97x on gemm1 at m=4); wide-N is a **loss at m=4**. Only the barrier-free variants gain a little (8x128+skip 1.05x/1.06x, 8x16 1.04x, 8x64-nobar 1.04x/1.03x). The code is retained so the negative result is reproducible, not because a default should change. Also in this commit: `avg_m` now rounds up — a truncating divide reported 0 rows/expert whenever `total_m < n_experts`, which is every decode step; latent only, since 0 and 1 both take the `avg_m <= 8` branch.

**Contiguity of `topk_ids` is deliberately not part of the `M == 1` gate.** The fused branch copies a strided `topk_ids` contiguous first; `prepare_moe_input` indexes it linearly over `numel` and reads the stride holes. Adding `is_contiguous()` to the gate would route strided ids *back into* the faulting path — `prepare_moe_input` indexes `topk_ids` linearly over `numel` (TODO at `python/sgl_kernel/moe.py:901`) and `src/sycl/MoEPrepareInputs.cpp:239` does an unbounded device-scope atomic RMW on `atomic_buffer_[expert]`. Both are pre-existing, neither is introduced or fixed here, and serving is unaffected because `topk` returns contiguous int32 ids.

## Accuracy Tests

Bit-exact, `torch.equal` rather than a tolerance, at every point measured:

- `7460aac`: bit-exact (`torch.equal` on the int16 view) at M=1 and M=1024, before vs after, plus cache-hit, version-invalidation and inference-tensor checks.
- `35e368b`: bit-exact against the four-kernel path at m=1 under the shipped gate, and at m=1, 2, 4, 8, 14 with the gate opened. **No bit-exactness claim at m=16 or 64** — `_should_use_small_moe_prepare` declines both (element cap 64x2880 = 184320 > 163840; route cap 256 > 64), so the fused kernel never ran there and the comparison would be the four-kernel path against itself. Path selection is read from captured kernel names, not from the gate expression.

pytest's own summary lines, on the built branch:

```
tests/test_norm.py                     823 passed,   16 skipped in  10.12s
tests/test_swiglu_with_alpha_limit.py 1010 passed              in   7.31s
tests/test_moe_prepare_input.py       4914 passed, 6623 skipped in 104.41s
```

**0 failed.**

**Build:** `pip install -v .` — 49 min 55 s, zero compiler errors, `CMPLR_ROOT` pinned to oneAPI 2026.1, `CMAKE_BUILD_PARALLEL_LEVEL=24`. Verified at runtime that the *installed* `site-packages/sgl_kernel/moe.py` carries both features (`_bias_as_float32`, the `M == 1` gate) — `.so` size is not evidence of what compiled in.

## Benchmark and Profiling Results

**Configuration for every figure below:** `openai/gpt-oss-120b`, Intel Arc Pro B60, **TP=4**, `--attention-backend intel_xpu`, MXFP4 MoE weights (W4A16), bf16 activations and bf16 KV cache, `bench_serving` random 4096-in / 1024-out, **concurrency 1**, `ZE_AFFINITY_MASK=4,5,6,7`. **n=1 per arm.**

### The per-patch result — 144 → 36 bookkeeping launches/step

The two affected buckets, as an explicit bridge:

| Bucket | launches/step | device self-XPU (ms/step) |
|---|---:|---:|
| `bookkeeping` (3 kernel families) | 108 | 0.1874 |
| `ScatterTokensToExperts` (carried in dispatch/combine) | 36 | 0.1237 |
| **pool removed** | **144** | **0.3111** |
| replaced by `PrepareMoeInputSmall` | 36 | 0.1207 |
| **measured** | **−108** | **−0.190** |

**0.3111 ms/step is the entire pool these two buckets contain — no figure above it is possible for this lever, and none is claimed.**

Two independent bases agree in sign and rank: the microbench predicted **−4.14 µs/layer (−0.149 ms/step)**; in-trace is **−5.29 µs/layer (−0.190 ms/step)**, larger because the trace basis includes `ScatterTokensToExperts`.

**`PrepareMoeInputSmall` is launch-bound and is not a future optimization target.** 0.1207 ÷ 36 = **3.35 µs/launch**, which sits on the ~0.9–3 µs launch floor. It is at its limit; ranking it for further optimization would be a misreading.

### Amdahl ceiling

The full P0/P1 subtotal of **−0.814 ms/step** of device self-time is **1.2%** of measured eager TPOT (68.25 ms/step). **Anything larger than that is arithmetic error, not a result.** The subtotal is device time; it is not asserted to carry into wall clock.

### Campaign-level context — decode buckets, v4 vs v5

> ⚠ **This table is campaign-level context, not an A/B of this patch.** Between the two campaigns `sgl-kernel-xpu` moved by its own upstream merge **plus PR 469**, and `sglang` moved by a **687-commit** merge. Reproducibility is not attributability: none of these deltas may be attributed to the four commits in this PR. The only per-patch attribution in this PR is the launch-count bridge above and the **−0.190 ms/step** it isolates.

| Bucket | v4 | v5 | delta |
|---|---:|---:|---:|
| Comms | 60.694 | 50.557 | −10.137 ⚠ unquotable — comms swings 55% across the four rank files inside one trace |
| MoE | 3.526 | 3.247 | −0.279 |
| GEMM | 2.421 | 2.218 | −0.203 |
| Norms + RoPE + KV + elementwise | 1.212 | 0.594 | −0.618 |
| FMHA | 0.380 | 0.361 | −0.019 |
| **Total** | **68.233** | **56.977** | **−11.256** ⚠ comms-dominated |
| **Total, comms excluded** | **7.539** | **6.420** | **−1.119** |

### Reproduce

```bash
conda activate sgl-xpu-d
export PATH=/opt/intel/oneapi/compiler/2026.1/bin:$PATH        # 2026.1 must precede 2025.3
export LD_LIBRARY_PATH=/opt/intel/oneapi/compiler/2026.1/lib:$LD_LIBRARY_PATH
export SGLANG_USE_SGL_XPU=1
export ZE_AFFINITY_MASK=4,5,6,7

# --- build sgl-kernel-xpu (CMPLR_ROOT pinned; PATH is ignored by the kernel build) ---
export CMAKE_BUILD_PARALLEL_LEVEL=24
export CMPLR_ROOT=/opt/intel/oneapi/compiler/2026.1
pip install -v .

# --- correctness ---
pytest -q tests/test_norm.py tests/test_swiglu_with_alpha_limit.py tests/test_moe_prepare_input.py

# --- microbenches for the two levers ---
python benchmark/bench_moe_dispatch_casts.py          # lever 1: the bf16->fp32 bias promotion
python benchmark/bench_moe_w4a16_policy_sweep.py      # the refuted wide-N hypothesis (SGL_W4A16_POLICY_ID)

# --- serving trace that produced the launch counts ---
cd /root   # avoid a local sglang/ dir shadowing the package
python3 -m sglang.launch_server \
  --model-path openai/gpt-oss-120b --served-model-name openai/gpt-oss-120b \
  --device xpu --tp 4 --attention-backend intel_xpu \
  --mem-fraction-static 0.8 --context-length 8192 --max-total-tokens 32768 \
  --max-running-requests 8 --disable-radix-cache --trust-remote-code \
  --reasoning-parser gpt-oss --tool-call-parser gpt-oss \
  --host 127.0.0.1 --port 8000 --chunked-prefill-size -1 --max-prefill-tokens 8192 \
  --model-loader-extra-config '{"enable_multithread_load": false}' \
  --weight-loader-disable-mmap
# profiled arm additionally sets SGLANG_TORCH_PROFILER_DIR before launch

python3 -m sglang.bench_serving \
  --model openai/gpt-oss-120b --backend sglang --host 127.0.0.1 --port 8000 \
  --warmup-requests 1 --request-rate inf \
  --dataset-name random --random-input-len 4096 --random-output-len 1024 \
  --random-range-ratio 1.0 --num-prompts 1 --max-concurrency 1 \
  --profile --profile-activities XPU --profile-num-steps 10
```

## Checklist

- [x] Correctness: bit-exact against the pre-change path at every measured point; 6747 passed / 6639 skipped / **0 failed** across the three named test files.
- [x] Builds clean with oneAPI 2026.1, zero compiler errors; presence of both features verified in the *installed* module at runtime.
- [x] No default behaviour change from `c8f1f4a` — the new policies are unreachable from `select_w4a16_policy_id()`.
- [x] The `M == 1` gate is exact equality and is justified by the m>=2 regression measurement in *Motivation*.
- [x] Every number in this body carries the command that produced it.
- [x] No TPOT, end-to-end or speedup claim; no extrapolation beyond the single measured configuration.
- [x] Pre-existing `topk_ids` stride / unbounded-atomic hazard disclosed in *Modifications*, and is why `is_contiguous()` was kept **out** of the new gate.
- [x] Rebased onto current `main` (`ecddde5`): the branch carries **only the four commits above**, and the six files are byte-identical to the tree that was built and tested.
- [x] `pre-commit run --all-files` clean at this head.


src/sycl/GroupGemmW4A16Xe20.cpp
```diff
@@ -134,7 +142,32 @@ at::Tensor& w4a16_atomic_counter(const torch::TensorOptions& options, c10::xpu::
...
+
+// Debug tile override for the policy A/B in
+// benchmark/bench_moe_w4a16_policy_sweep.py: SGL_W4A16_POLICY_ID=<id> forces one
+// compiled policy for every call, which is the only way to time a tile that
+// select_w4a16_policy_id() would never choose. Read once per process; unset or
+// out of range means no override. Not a production knob -- an m=512 expert on the
+// 8-row tile is correct but very slow.
+int w4a16_policy_override() {
+  static const int forced = [] {
+    const char* env = std::getenv("SGL_W4A16_POLICY_ID");
```
airMeng, line 157: I'd prefer some heuristic judgements not another environmental variables which might confuse the users

python/sgl_kernel/moe.py
```diff
@@ -412,6 +412,79 @@ def _get_moe_ws(
...
+    reuses a tensor — would otherwise be served a stale promotion. Rebinding
+    `layer.w13_weight_bias.data = ...` instead produces a new tensor object and
+    misses the cache outright, which is also correct. The cache is bounded so a
+    caller that churns short-lived bias tensors cannot grow it without limit.
+    """
+    key = id(bias)
+    version = _version_or_none(bias)
+    hit = _moe_bias_f32_cache.get(key)
+    if hit is not None and hit[0] is bias and hit[1] == version:
+        return hit[2]
```
YangKai0616, line 479: This cache can return a stale bias after an online weight update. SGLang’s default loader uses `param.data.copy_(...)`, which changes the values without changing the Tensor identity or `_version`; inference tensors have no version counter either. I reproduced a cache hit returning the old fp32 value after `.data.copy_`.

Please move the one-time promotion to SGLang’s weight-loading lifecycle and refresh it on updates, or provide a reliable invalidation mechanism.

src/GroupGemmW4A16Xe20.cmake
```diff
@@ -18,16 +18,27 @@ endfunction()
 # layouts and have separate N-tail-skip instantiations, because that bit changes
 # device code. select_w4a16_policy_id() chooses from this table per call.
 # group_size (32/64/128/256) is compiled into every unit as a runtime branch,
-# so it does not multiply the instance count. Total: 7 policies x 2
-# (int4/mxfp4) x 2 (bf16/fp16 activation) = 28 units.
+# so it does not multiply the instance count. Total: 14 policies x 2
+# (int4/mxfp4) x 2 (bf16/fp16 activation) = 56 units.
```
YangKai0616, line 22: These 7 policies are unreachable by the default selector, but they double the AOT instantiations from 28 to 56. Could we move the tile sweep and override to a separate experimental PR, then add only policies with demonstrated, repeatable gains to the production AOT/JIT dispatch?

PR #497, opened 2026-09-18
Add fused scale-residual + LayerNorm + scale-shift SYCL kernel for XPU

ckvermaAI:
Diffusion DiT blocks (Wan, Qwen-Image) apply the same sequence around every attention and MLP branch:

  residual_output = residual + x * gate
  out             = layer_norm(residual_output)[* weight + bias] * (1 + scale) + shift

Run eagerly this is five passes over the activations. This fuses it into one, porting the Triton reference at
sglang/python/sglang/kernels/ops/diffusion/norm/scale_residual_norm_scale_shift_triton.py so XPU gets a compiled torch op instead of relying on Triton. Both `out` and `residual_output` are returned, and the normalization statistics are taken from the unrounded fp32 residual sum, matching the Triton kernel bit-closely.

Plain SYCL, not CUTLASS: the op is a streaming elementwise pass plus one row-wise reduction with no matrix product, so sycl-tla's collective mainloops and tile schedulers have nothing to accelerate. This matches how every other norm/elementwise kernel here is written (RMSNorm.cpp, Norm.h, SiluAndMulClamp.cpp); in this tree CUTLASS is used only by the GEMM and attention kernels.

Kernel design:
- One work-group (256 threads) per row. The row is held in a per-thread fp32 register array across the two reduction passes (mean, then variance), so it is read from global memory once. That array is what bounds hidden to 8192, the same limit the Triton reference sets.
- Vector width (8/4/2/1) is chosen per call on the host from the hidden size and the base address of every vector-loaded operand, then dispatched to a template instantiation, so a misaligned-but-contiguous slice narrows the width instead of faulting.
- Specialized on the (activation dtype, parameter dtype) pair, so fp32 LayerNorm parameters work against bf16/fp16 activations.
- Gate mode, affine, and scalar-vs-vector modulation are runtime flags rather than template parameters. They are work-group-uniform branches in a bandwidth-bound kernel, and keeping them dynamic avoids 32x the instantiations.

Covers every operand form the Triton op accepts: ungated / per-token [1, 1, hidden] / per-frame [1, num_frames, 1, hidden] gate, scalar or per-channel scale and shift, and LayerNorm with or without affine parameters. 145 tests in tests/test_diffusion_scale_residual_norm.py sweep hidden 64..8192 across all four vector widths, sequence lengths 1/17/1024, fp16/bf16/fp32, fp32-parameter mixes, misaligned slices, and 11 unsupported operand sets that `can_use_*` must reject and the kernel must raise on.

Measured on one B60, bf16, per-token gate, against a torch eager equivalent (stable to +-0.001 ms over three runs):

  S=1024   H=1536   0.053 ms  238 GB/s   1.10x (launch-bound)
  S=9216   H=3072   0.578 ms  392 GB/s   3.11x
  S=32760  H=5120   3.414 ms  393 GB/s   3.31x

394 GB/s is roughly 86% of the card's ~456 GB/s peak, so the kernel is close to bandwidth-bound optimal.


ckvermaAI:
@pralay-das Please review this PR

src/sycl/ScaleResidualNormScaleShift.cpp
```diff
@@ -0,0 +1,447 @@
...
+  TORCH_CHECK(
+      x.scalar_type() == at::ScalarType::Float || x.scalar_type() == at::ScalarType::Half ||
+          x.scalar_type() == at::ScalarType::BFloat16,
+      "fused_scale_residual_norm_scale_shift: x must be float32, float16 or bfloat16, got ",
+      x.scalar_type());
+  TORCH_CHECK(
+      x.dim() == 3 && x.size(0) == 1,
+      "fused_scale_residual_norm_scale_shift: x must be [1, seq_len, hidden], got ",
+      x.sizes());
+  TORCH_CHECK(x.is_contiguous(), "fused_scale_residual_norm_scale_shift: x must be contiguous");
```
pralay-das: use `CHECK_INPUT` instead of check separately xpu device and contiguous.

src/sycl/ScaleResidualNormScaleShift.cpp
```diff
@@ -0,0 +1,447 @@
...
+  TORCH_CHECK(
```
pralay-das, line 323: use `CHECK_INPUT`.

src/sycl/ScaleResidualNormScaleShift.cpp
```diff
@@ -0,0 +1,447 @@
...
+  const int64_t hidden = x.size(2);
+  TORCH_CHECK(
+      hidden <= kSrnssMaxHidden,
+      "fused_scale_residual_norm_scale_shift: hidden must be <= ",
+      kSrnssMaxHidden,
+      ", got ",
+      hidden);
+  TORCH_CHECK(
+      weight.has_value() == bias.has_value(),
+      "fused_scale_residual_norm_scale_shift: weight and bias must both be given or both be absent");
```
pralay-das: if this is a strict condition, then is it a mandatory to create both as an optional tensor?

src/sycl/ScaleResidualNormScaleShift.cpp
```diff
@@ -0,0 +1,447 @@
...
+  } else if (act_dtype == at::ScalarType::BFloat16) {
+    dispatch_param_dtype<::sycl::ext::oneapi::bfloat16>(queue, param_dtype, a);
+  } else {
+    TORCH_CHECK(false, "fused_scale_residual_norm_scale_shift: unsupported dtype ", act_dtype);
+  }
+}
+
+void check_param(const torch::Tensor& t, const torch::Tensor& x, at::ScalarType param_dtype, const char* name) {
+  TORCH_CHECK(t.device() == x.device(), "fused_scale_residual_norm_scale_shift: ", name, " must be on x's device");
+  TORCH_CHECK(t.is_contiguous(), "fused_scale_residual_norm_scale_shift: ", name, " must be contiguous");
```
pralay-das: same here, use `CHECK_INPUT`

src/sycl/ScaleResidualNormScaleShift.cpp
```diff
@@ -0,0 +1,447 @@
...
+      gate_row_div,
+      gate_row_stride,
+      scale_is_vec,
+      shift_is_vec,
+      select_vec_size(hidden, vector_operands),
+      static_cast<float>(eps)};
+
+  auto stream = at::xpu::getCurrentXPUStream();
+  auto& queue = stream.queue();
+  dispatch_dtypes(queue, x.scalar_type(), param_dtype, args);
```
pralay-das: templatize the dispatch call based on the dtype.

pralay-das (commented):
add the benchmark file

src/sycl/ScaleResidualNormScaleShift.cpp
```diff
@@ -0,0 +1,447 @@
...
+
+    float thread_var = 0.0f;
+#pragma unroll
+    for (int j = 0; j < kMaxVecsPerThread; ++j) {
+      const uint32_t i = tid + static_cast<uint32_t>(j) * kSrnssThreadsPerBlock;
+      if (i < num_vecs) {
+#pragma unroll
+        for (int v = 0; v < kVecSize; ++v) {
+          const float centered = row_vals[j][v] - mean;
+          row_vals[j][v] = centered;
```
pralay-das, line 144: use vector load and store.

src/sycl/ScaleResidualNormScaleShift.cpp
```diff
@@ -0,0 +1,447 @@
...
+        const Vec rv = residual_vec[i];
+        PVec gv;
+        if (gate_vec != nullptr) gv = gate_vec[i];
+        Vec res_out;
+#pragma unroll
+        for (int v = 0; v < kVecSize; ++v) {
+          float scaled = static_cast<float>(xv[v]);
+          if (gate_vec != nullptr) scaled *= static_cast<float>(gv[v]);
+          const float val = static_cast<float>(rv[v]) + scaled;
+          row_vals[j][v] = val;
```
pralay-das, line 125: same here, use vec load and store.

PR #501, opened 2026-09-21
perf: add CUTLASS_SYCL_PROFILING_ENABLED-gated per-kernel perf reporting

arathi-hlab:
## Motivation

Today there is no uniform way to get per-kernel GPU time / bandwidth / TFLOPS out of the XPU kernels. this PR generalizes that pattern into a single reusable helper and wires it into 34 SYCL kernel entry points, so any kernel can be profiled from a single build flag without touching kernel code.

## Modifications

- **New `src/sycl/SGLKernelPerf.h`**: exposes `::sglkernel::report_kernel_perf(op_name, queue, timer, bytes, flops)`, which waits on the queue, reads the CUTLASS `GPU_Clock`, and prints one line:
  `<op> perf(gpu_clock): time=X ms (Y us), bandwidth=Z GB/s, compute=W TFLOPS (V GFLOPS)`
  with an `unavailable (reported 0)` fallback when the measured interval is non-positive. The whole header body is inside `#if defined(CUTLASS_SYCL_PROFILING_ENABLED)`.
- **`CMakeLists.txt`**: propagate `CUTLASS_SYCL_PROFILING_ENABLED` as a compile definition when the CMake option is set, so the instrumentation is opt-in via `-DCUTLASS_SYCL_PROFILING_ENABLED=ON`.
- **`src/sycl/SYCLHelpers.h`**: `sycl_kernel_submit()` overloads now return the `sycl::event` instead of discarding it (previously `void`). This is source-compatible with all existing call sites and makes the submitted work observable for timing.
- **34 kernel entry points instrumented** with the same guarded wrap (timer start before the launch, `report_kernel_perf()` with op-specific BYTES/FLOPS formulas after):
  - MLA / attention: `mla_sparse_decode`, `mla_sparse_prefill`, `flash_attention`, `KVCache`, `Fp8MqaLogitsXe20`, `gdn_attn_interface_impl`
  - top-k / sampling: `BiasedTopK`, `HashTopK`, `TopKSoftMax`, `TopKTopPSamplingFromProbs`, `TopKTransform`
  - hicache / compression: `CompressPlan`, `FlashCompress4`, `FlashCompress128`, `HCHead`, `HCPost`, `HCPreFuse`, `HCPreGemmSqrSum`, `MHCFusedPostPre`
  - MoE: `MoEAlign`, `MoEPrepareInputs`, `MoE_fused_gate`, `MoE_sum_reduce`
  - fused RoPE / QK-norm: `FusedNormRopeStore`, `FusedQIndexerRopeHadamardQuant`, `FusedQKNormRope`, `FusedQKRope`, `MRope`
  - grouped GEMM: `GroupGemmW4A16Xe20`, `GroupGemmW8A16Xe20`, `TripleOps`
  - norm / activation: `RMSNorm`, `SiluAndMulClamp`, `SwigluAlphaLimit`

Default builds do not define `CUTLASS_SYCL_PROFILING_ENABLED`, so no timer, no `queue.wait()`, and no printf are compiled in — release behavior and performance are unchanged.

## Usage

```bash
cmake -B build -DCUTLASS_SYCL_PROFILING_ENABLED=ON ...
```

Then every instrumented op prints its own perf line on each launch.

## Checklist

- [x] Instrumentation is fully compile-gated; default (non-profiling) builds are unaffected.
- [x] `pre-commit run` passes on all touched files.


arathi-hlab:
Local run result for basic configs
**Sparse-MLA prefill smoke**

[quick-prefill] MIN config: s_q=16 h_q=8 topk=8 d_qk=512
[quick-prefill] CPU inputs built in 0.05s
[quick-prefill] H2D transfer in 10.10s
[quick-prefill] calling flash_mla_sparse_fwd (1x)
**flash_mla_sparse_prefill perf(host_wall): time=9249.567098000 ms (9249567.098 us), bandwidth=0.000043 GB/s, compute=0.000000 TFLOPS (0.000 GFLOPS)**
[quick-prefill] wall: 9.27s
[quick-prefill] out.shape=(16, 8, 512) dtype=torch.bfloat16

**Sparse-MLA decode smoke**

[quick-decode] MIN config: batch=1 topk=8 H=16 num_pages=8 page_size=256
[quick-decode] CPU inputs built in 0.01s
[quick-decode] H2D transfer in 8.11s
[quick-decode] calling flash_mla_with_kvcache (1x)
f**lash_mla_sparse_decode perf(host_wall): time=4281.980074000 ms (4281980.074 us), bandwidth=0.000009 GB/s, compute=0.000000 TFLOPS (0.000 GFLOPS)**
[quick-decode] wall: 4.31s

sunjiweiswift (approved):

pralay-das:
**suggestion:** 
fold the repeated profiling boilerplate into the header.

Every instrumented file repeats two compile-gated blocks — the `#if … #include … #endif` and the `GPU_Clock timer; timer.start();` pair. Both can be consolidated so call sites carry no `#if` and the timer/report collapses to one line.

1. _Make the header self-guarding_ — include `GPU_Clock.hpp` internally (inside the enabled branch), keep the body under `#if defined(CUTLASS_SYCL_PROFILING_ENABLED),` add `#else` no-op. Call sites then just:
```
#include "SGLKernelPerf.h"   // no #if at call sites
```
2. _Add a RAII scope macro for the timer/report pair:_
```
#if defined(CUTLASS_SYCL_PROFILING_ENABLED)
#include <cutlass/util/GPU_Clock.hpp>
namespace sglkernel {
struct KernelPerfScope {
  const char* op; sycl::queue& q; double bytes, flops; GPU_Clock timer;
  KernelPerfScope(const char* o, sycl::queue& q_, double b, double f)
      : op(o), q(q_), bytes(b), flops(f) { timer.start(); }
  ~KernelPerfScope() { report_kernel_perf(op, q, timer, bytes, flops); }
};
}  // namespace sglkernel
#define SGL_PERF_CAT_(a, b) a##b
#define SGL_PERF_CAT(a, b)  SGL_PERF_CAT_(a, b)
#define SGL_KERNEL_PERF_SCOPE(op, q, b, f) \
  ::sglkernel::KernelPerfScope SGL_PERF_CAT(sgl_perf_scope_, __LINE__)((op), (q), (b), (f))
#else
#define SGL_KERNEL_PERF_SCOPE(op, q, b, f) ((void)0)
#endif
```
Each kernel reduces to one line before the launch:
```
SGL_KERNEL_PERF_SCOPE("flash_attention", queue, bytes, flops);
sycl_kernel_submit(...);
```
Correctness notes:

- Two-level concat is required. `foo##__LINE__` does not expand `__LINE__`, and any identifier containing __ is reserved to the implementation — hence the `SGL_PERF_CAT` indirection and the non-reserved `sgl_perf_scope_` prefix.

- Scope-exit timing semantics. `GPU_Clock::stop()` (invoked inside milliseconds()/seconds()) records the queue's last event at destruction. So the guard measures correctly only when the target submit is the last queue operation in its scope. Use `SGL_KERNEL_PERF_SCOPE` for the submit-then-return kernels (the majority); for entry points that submit multiple kernels or do post-submit queue work, keep the explicit `report_kernel_perf()` call right after the relevant submit. Keep that function public for this reason.

arathi-hlab:
@airMeng  @pralay-das  @mkumargarg  @mingfeima please review once

arathi-hlab:
@pralay-das thanks for the suggestion — applied in e982faa.

**Header** (`src/sycl/SGLKernelPerf.h`): added `sglkernel::KernelPerfScope` (RAII: starts `GPU_Clock` at construction, calls `report_kernel_perf()` at destruction) and the `SGL_KERNEL_PERF_SCOPE(op, queue, bytes, flops)` macro. Two-level `SGL_PERF_CAT` for `__LINE__` expansion + non-reserved `sgl_perf_scope_` prefix, as you noted. `#else` branch is a no-op that consumes each argument via `(void)...` so call-site locals used only for `bytes`/`flops` don't trip `-Wunused` in non-profiling builds. `report_kernel_perf()` stays public for the multi-submit case per your correctness note.

**Migrated 23 single-submit call sites**: `flash_attention`, `mla_sparse_{decode,prefill}`, `KVCache`, `BiasedTopK`, `TopKSoftMax`, `SwigluAlphaLimit`, `SiluAndMulClamp`, `MRope`, `MoE_sum_reduce`, `MoE_fused_gate`, `MoEAlign`, `HCPreGemmSqrSum`, `HCPreFuse`, `HCPost`, `HCHead`, `HashTopK`, `GroupGemmW{4,8}A16Xe20`, `gdn_attn/gdn_attn_interface_impl`, `FusedQIndexerRopeHadamardQuant`, `FusedNormRopeStore`, `TopKTopPSamplingFromProbs`. Each collapses two `#if` blocks + a manual timer into one line before the launch.

**Left explicit** (per your submit-then-return guidance): `MHCFusedPostPre` (report inside a conditional branch), and the 8 multi-submit files: `MoEPrepareInputs`, `FusedQKNormRope` (4 reports each), `TopKTransform`, `FusedQKRope`, `Fp8MqaLogitsXe20`, `FlashCompress{4,128}`, `CompressPlan` (2 reports each).

**HCHead note**: `y` is allocated inside the `SYCL_DISPATCH` lambda, so `y.numel()` / `y.element_size()` aren't available at scope-construction time. Since `launch_fused_hc_head` does `at::empty({T, D}, x.options())`, `y_bytes` equals `T * D * x.element_size()` — used that equivalent expression so the scope brackets the launch.

**Follow-up**: the 28 kernels added in 7ca8830 (`EmbeddingLoraAFwd`, `Rope`, `mla_decode`, `mla_prefill`, `awq_dequantize`, per-token/tensor quant, etc.) still use the old two-`#if` pattern. Happy to do a second pass in a follow-up commit if you'd like.

Net: 24 files, +293/-383 lines. No functional / print-shape change (perf line still fires just before the enclosing function returns; the scope destructor runs at end of function scope).

arathi-hlab:
@pralay-das applied in c55a483 — **52 files, +510 / −861**.

**Header** (`SGLKernelPerf.h`): added `sglkernel::KernelPerfScope` + `SGL_KERNEL_PERF_SCOPE(op, queue, bytes, flops)` macro (two-level `SGL_PERF_CAT` for `__LINE__`, `sgl_perf_scope_` prefix). Self-guarding include, `#else` no-op consumes each arg via `(void)...`. `report_kernel_perf()` kept public for the multi-submit case.

**47 single-submit sites migrated** to the scope macro.

**13 files kept explicit** per your submit-then-return note (12 multi-submit + `MHCFusedPostPre` conditional-report) — those still got the top-of-file `#if…#include…#endif` wrap collapsed to a plain `#include "SGLKernelPerf.h"`.

**Non-mechanical substitutions** (bytes/flops computed pre-launch):
- `HCHead`: `y` allocated inside `SYCL_DISPATCH` — used `T * D * x.element_size()` (launcher does `at::empty({T, D}, x.options())`).
- `Rope`: `std::get<0/1>(result)` replaced with `query`/`key` (same shape/dtype in both branches).
- `Hadamard`: trailing `reshape` may submit a contiguous copy after the launch — wrapped the launch in a nested `{ ... }` so the scope destructs before it.

pralay-das (approved):
still some refactoring is pending, check in later PR, otherwise LGTM.

PR #503, opened 2026-09-21
Enable MLA Sparse for XE_35 

pralay-das:


benchmark/bench_flash_mla_sparse_fwd.py
```diff
@@ -191,25 +231,53 @@ def effective_bytes(s_q, h_q, topk, d_qk):
...
+    PrefillConfig(2048, 16, 512, 512, S_KV),
+    PrefillConfig(2048, 128, 512, 512, S_KV),
+    PrefillConfig(512, 16, 2048, 576, S_KV),
+    PrefillConfig(512, 32, 2048, 576, S_KV),
+    PrefillConfig(512, 128, 2048, 576, S_KV),
+    PrefillConfig(2048, 16, 512, 576, S_KV),
+    PrefillConfig(2048, 128, 512, 576, S_KV),
+]
+
+# ---- PREFILL-stage for DEEPSEEK V4 flash ----
```
jayachandranb-ai, line 257: is it ok to add all these tests to CI ( will they cause CI timeout?)

benchmark/bench_flash_mla_sparse_fwd.py
```diff
@@ -218,47 +286,46 @@ def effective_bytes(s_q, h_q, topk, d_qk):
...
+        q, kv, indices, topk_length = build_inputs(
+            cfg.s_q,
+            cfg.h_q,
+            cfg.topk,
+            cfg.d_qk,
+            cfg.s_kv,
+            use_topk_length=cfg.use_topk_length,
+            device=device,
+        )
+        sm_scale = cfg.d_qk**-0.5
```
jayachandranb-ai, line 299: for dsv4 , it should be fine to take d_qk directly to calculate sm_scale, but for dsv3.2, we need to take the actual dqk (192). Good to have that distinction though sm_scale is passed to the kernel from outside

benchmark/bench_flash_mla_sparse_fwd.py
```diff
@@ -155,26 +168,53 @@ def flash_mla_sparse_prefill_triton(
 
     invalid_mask = (flat_indices < 0) | (flat_indices >= s_kv)
 
-    gathered_kv = _gather_dense(kv, flat_indices, d_qk)
-    out, lse = _compute_attention(
-        q, gathered_kv, invalid_mask, sm_scale, d_v, attn_sink
-    )
+    outs, lses = [], []
+    for start in range(0, s_q, _REF_CHUNK_SQ):
```
jayachandranb-ai, line 172: why was the reference code chunked?

tests/_xpu.py
```diff
@@ -0,0 +1,65 @@
+"""The single XPU-architecture probe for the test suite.
```
jayachandranb-ai, line 1: looks like this is a copy of the code under bench.
let us ee if it can be commonized going forward(not necessarily in initial PR)

benchmark/bench_flash_mla_sparse_fwd.py
pralay-das: modified.

benchmark/bench_flash_mla_sparse_fwd.py
pralay-das: without chunking it is giving OOM issue.

tests/_xpu.py
pralay-das: sure, we will take care in next PR.

benchmark/bench_flash_mla_sparse_fwd.py
pralay-das: these test configurations are used in real workload, so I think it is good to have it here to catch the regressions in early.

PR #504, opened 2026-09-21
[XPU] Fuse QK into the fp8 paged MQA logits triton kernel

unrahul:
## Motivation

`fp8_paged_mqa_logits_triton` computes its QK product as an eager fp32
`torch.einsum` outside the kernel. Xe2 has no fp32 DPAS path, so that product
never reaches the matrix engines. The kernel also materialises the full score
tensor between its two stages.

## Modifications

The paged gather, dequantization, QK product and the relu / weight /
head-reduction / scale / mask pipeline now run in one kernel. `q` is loaded once
per program as `[HEAD_DIM, BLOCK_H]` and reused across a loop over pages;
`BLOCK_SZ` equals the KV page size. The product is a bf16 `tl.dot` with an fp32
accumulator. `_score_relu_weight_scale_kernel` and the chunking loop are removed,
both module-private. The public signature is unchanged.

Pages per program keeps the grid at `_MIN_PROGRAMS` or above, rounded down to a
power of two, so the `tl.constexpr` value set stays bounded and does not compile
new binaries during serving.

## Accuracy Tests

`tests/test_fp8_paged_mqa_logits.py`, already in the per-commit suite, against
the fp32 reference at `atol=rtol=1e-3`. 14 tests pass, the 9 pre-existing ones
unchanged. Five are added: the existing cases all cap sequence length at 256,
four pages, so none of them enters the loop. The new ones cover
multi-page-per-program with an odd page count, and the tile selection over a grid
of batch and page counts.

## Benchmarking and Profiling

One card, `--triton-B 64 --H 64 --D 128`, page size 64:

| keys | before | after | speedup | after TFLOPS |
|---:|---:|---:|---:|---:|
| 1,024 | 381.9 | 191.1 | 2.00x | 5.62 |
| 4,096 | 1,203.6 | 547.6 | 2.20x | 7.84 |
| 16,384 | 4,679.0 | 1,713.0 | 2.73x | 10.03 |
| 32,768 | 9,349.1 | 3,335.5 | 2.80x | 10.30 |

Both arms in the same container on the same card, alternating, three warmups.

`benchmark/bench_fp8_mqa_logits.py` paged case only covered the SYCL op at page size 128. `make_kv_cache` was a Python double loop with a `struct.pack` per token, minutes at these page counts, and is now vectorised.


unrahul:
@mingfeima @airMeng @chunyuan-w @sunjiweiswift review please. I cannot set reviewers directly on this repo.

PR #505, opened 2026-09-21
[XPU] Add the DeepSeek V4 expand_prefill_causally attention metadata op

unrahul:
## Motivation

The DeepSeek V4 attention path needs `expand_prefill_causally`, which expands
per-request sequence lengths into per-token causal lengths and repeats the
request pool indices alongside them. On XPU neither existing arm is usable: the
triton kernel for this step does not compile under triton-xpu 3.7.2, and the
eager torch reference is not numerically equivalent to the kernel it stands in
for. This registers a SYCL implementation so the framework has a correct arm to
dispatch to.

## Modifications

Registers `dsv4_expand_prefill_causally_out` with a SYCL implementation in
`src/sycl/DSV4ExpandPrefill.cpp`, the schema and dispatch in
`src/torch_extension_sycl.cc` and `include/sgl_kernel_ops.h`, and a Python
wrapper in `python/sgl_kernel/dsv4_prefill.py`.

The op is out-of-place in the sense that both outputs are caller-allocated, so it
holds no state and performs no host sync. That matters because the caller runs it
under an active command graph, where a device to host sync is illegal.

`padded_num_tokens` lets the caller size the outputs to the padded token count.
The tail beyond `num_tokens` is filled with a causal length of 1 and the last
request's pool index, so a graph captured at a fixed width has defined values
across the whole output rather than whatever the buffer held.

Purely additive: 708 insertions, no deletions, no existing behaviour touched.

## Accuracy Tests

`tests/test_dsv4_attn_metadata.py`, registered in both the per-commit suite
(`tests/run_suite.py`) and the nightly suite (`tests/run_nightly_suite.py`).
Seven test functions, several parametrized, against a Python reference:

* the reference match over a grid of batch sizes and sequence lengths
* clamped spans and the padding tail beyond `num_tokens`
* an empty batch, which must be a no-op
* two negative controls, one perturbing the causal base and one shifting
  `extend_start_loc`, so the equality assertions cannot pass vacuously
* argument validation, one case per rejected mutation
* graph replay, asserting the op rereads its inputs rather than caching them

## Benchmarking and Profiling

None. This adds an operator that had no working XPU arm, so there is no before
number to compare against. The correctness case is the unit tests above.


unrahul:
@mingfeima @airMeng @chunyuan-w @sunjiweiswift review please. I cannot set reviewers directly on this repo.

PR #508, opened 2026-09-22
[softcap] Only compile the soft-cap kernel variant for head_dim 128/256

desmond-intel:
#491 added `Softcap` as a compile-time dispatch axis in every FMHA prefill/decode translation unit, which doubled the kernel instantiation count and noticeably increased the `build-and-test` "Rebuild sgl-kernel" step.

`attn_logit_softcapping` only exists in the Gemma-2 family, whose head dims are 128 and 256. This restricts the `Softcap=true` instantiation to those head dims (a shared `FMHA_DISPATCH_SOFTCAP` macro in `common.h`, selected per TU by `#if @HEAD_DIM@ == 128 || == 256`, else forcing `Softcap=false` + a `TORCH_CHECK`), mirroring the per-head-dim `#if` gating already used for sink (==64) and relative attention (==128). It also drops the dead `Softcap=true` variants on the fp8 KV path (no fp8 soft-cap model — same rationale as the existing `return_softmax_lse`/sink refusals) and on the rel_bias branches (the runners already reject softcap+rel_bias).

### Why head_dim 128/256

Only two architectures pass a nonzero attention softcap to the kernel:

| model | softcap | head_dim |
| --- | --- | --- |
| Gemma-2 | `attn_logit_softcapping` (50.0) | 128 (27b), 256 (2b/9b) |
| Grok-1 | `max(attn_logit_softcapping, 30.0)` (30.0) | 128 (`getattr(config, "head_dim", 128)`) |

Gemma-3/3n/4 hard-code `logit_cap=0.0` (the Gemma-3+ line dropped attention soft-capping), and no other model sets it. So the union of head dims that ever run softcap is {128, 256}. This is a feature-to-head-dim allowlist in the same style the file already uses (sink → 64, rel_bias → 128/512); anything outside it with `softcap>0` raises the `TORCH_CHECK` rather than silently dropping the cap, so adding a head dim later is a one-line change.

### Result

From a cold FMHA build: the heaviest objects (head_dim 512) drop 70–93 s each (e.g. `prefill_page_512_fp16` 261→185 s), while 128/256 are unchanged. Total FMHA compile CPU is down ~23%.

Verified: softcap at head_dim 128 is byte-identical to #491 (probe diff 1.08 / 3.91), and a non-gated head dim with `softcap>0` raises a clear error instead of silently dropping the cap. The softcap tests skip `softcap>0` outside the supported matrix.


Valentine233 (commented):
This is a bit complicated to define `FMHA_DISPATCH_SOFTCAP` in each `.cpp.in`. Could we modify like this?

In `common.h`:
```
#include <type_traits>
template <int HeadDim, bool IsFp8, typename Fn>
void dispatch_softcap(float softcap, Fn&& fn) {
  if constexpr (IsFp8) {
    TORCH_CHECK(
        softcap == 0.0f,
        "logit soft-cap is not supported with an fp8 KV cache");

    fn(std::false_type{});
  } else if constexpr (HeadDim == 128 || HeadDim == 256) {
    if (softcap != 0.0f) {
      fn(std::true_type{});
    } else {
      fn(std::false_type{});
    }
  } else {
    TORCH_CHECK(
        softcap == 0.0f,
        "logit soft-cap is only compiled for head_dim 128 and 256");

    fn(std::false_type{});
  }
}
```

In each `.cpp.in`:
```
dispatch_softcap<@HEAD_DIM@, false/true>(
    params.softcap, [&](auto softcap_tag) {
      constexpr bool Softcap = decltype(softcap_tag)::value;

      XXXConfig<
          ...,
          Softcap,
          ...>::run_paged(params);
    });
```

desmond-intel:
Thank you for the suggesting the much cleaner path. I've moved the logic into a `dispatch_softcap<HeadDim, IsFp8>` helper in `common.h` and call it with a lambda at each site as you proposed, so the per-file macro/`#if` is gone. The `if constexpr` branches still instantiate `Softcap=true` only for head_dim 128/256 on the non-fp8 path, so the build-time improvement is preserved.

I've also added negative tests covering the head-dim and fp8 soft-cap rejection paths. Rebuilt cleanly, and gemma-2-27b hellaswag acc_norm 0.55 is unchanged.

Valentine233 (approved):
LGTM. Please double-check the compiling time after the CI result comes out.

